PR #10, opened 2017-07-16
*.csproj changed to UTF-8 encoding

nlundby:
All *.csproj project files changed to UTF-8 encoding.  If an encoding declaration was used it was additionally removed. *.encrypt files were not touched.  After I switched the encoding I rebuilt the entire project and tested it using _runtests.sh_. All tests passed no errors.

Additionally, I received a good _git diff_ test once encoding had been changed to UTF-8. 

Lastly, I renamed 'chapter13' to 'Chapter13' and updated the EssentialCSharp.sln to reflect the capitalization.


MarkMichaelis:
This is all great Neal. Thanks so much for the help.

PR #19, opened 2017-07-24
Shared Code Restructure

nlundby:
### Deletions

- **app.config** files - Not all projects had an app.config file so I chose to delete them all. I figure we have them for all projects or no projects.  Is this the direction we want to take?
- Chapter **Program.cs** files - Not required when referencing the shared code.
- **AssemblyInfo.cs** - I couldn't find a use for this file.  It looks dated. Will anyone miss it?
- **enumtest.txt** - See above.
- **SharedCode Project** - The entire purpose of this.

### Shared Code

I reconciled _SharedCode Project_ files and the files in the _Shared_ directory.  It appears the Shared Code Project files were more current than the files in the _Shared_ so they took precedence. 

### ChapterMain()

All references changed to Main().  However, on a similar chain of thought, there does not seem to be a standard in the way the primary class is named, often it is the title of the listing section e.g. **Chapter01.Listing01_01 public class _HelloWorld_** otherwise it is named like in **Chapter01.Listing01_03 public class _Program_**. “Program” seems to be used the most... did we want to standardize this?

### EssentialCSharp.sln:

Reorganized and deleted appropriate references.  Removed global.json - will anyone miss it?

### *.csproj

Added linked files and appropriate references.   If a tuples reference was present it was retained.

Any pointers on how things could have been done better are welcome.  I tried to squash all my commits into one to make it cleaner, but it doesn't appear that worked.


MarkMichaelis:
This looks great Neal and the merge was easy. Thanks for doing this!!!

PR #38, opened 2018-10-19
V7.0 errata 9

kamonson:
Changes to support book errata 9 and 10, hours is not used in listing 9.7 degrees is.


MarkMichaelis:
It appears this has already been merged.

src/Chapter02/Chapter02.csproj
```diff
@@ -6,6 +6,9 @@
     <StartupObject>AddisonWesley.Michaelis.EssentialCSharp.Shared.Program</StartupObject> 
   </PropertyGroup>
   <Import Project="..\Versioning.targets" />
+  <ItemGroup>
```
adamskt: Looks like the project system got a little confused here, unless you actually want a code file to be in the project but not compiled.

COsborn2:
Changes incorporated

PR #51, opened 2019-06-25
Repair listing runner

COsborn2:
- Allow for user to type 2.1-2.6 to run a range of listings or 02.01-02.06
- Additionally allow user to type 2.1 to run the 2.1-2.6 listing
- Change type of argument[] from object[] to string[] to allow user input to be properly passed
- Update Readme.md to reflect changed behavior
- Remove Build.sln and update EssentialCSharp.sln to remove references to submodules. Neither solution will now reference the submodules and will instead use the Nuget package.


src/Shared/Program.cs
```diff
@@ -10,11 +11,17 @@ public class Program
     {
         public static void Main(string[] args)
         {
+            var assembly = Assembly.GetEntryAssembly();
+
+            string regexMatch = Regex.Match(assembly.FullName.Split(',')[0], "\\d{1,2}").Value;
```
Keboo: Is there a reason to use `FullName` and split on the comma rather than just use `Name`?

src/Shared/Program.cs
```diff
@@ -10,11 +11,17 @@ public class Program
...
+
+            int.TryParse(regexMatch, out int chapterNumber);
```
Keboo:
```suggestion
            int chapterNumber = int.Parse(regexMatch);
```

src/Shared/Program.cs
```diff
@@ -169,10 +175,27 @@ private static string ParseListingName(string listing, out string chapterName)
             for (int index = startPosition; index < chapterListing.Length; index++)
             {
                 if (index == startPosition && string.IsNullOrEmpty(chapterName)) chapterName = "Chapter" + chapterListing[index].PadLeft(2, '0');
-                listing += chapterListing[index].PadLeft(2, '0') + ".";
+                listing += chapterListing[index].PadLeft(2, '0') 
+                           + ((index+1 != chapterListing.Length) ? "." : "");
+            }
+
+            if (listing.Split('.').Length > 2)  // 02.01.02.06
```
Keboo: Perhaps something like this....
```cs
string[] parts = listing.Split('.');
if (parts.Length > 2)
{
    listing = $"({parts[0]}.{parts[1]}To{string.Join('.', parts.Skip(2))})";
}
```

src/Shared/Program.cs
```diff
@@ -63,7 +67,9 @@ public static void Main(string[] args)
                 }
                 else
                 {
-                    method.Invoke(null, arguments);
+                    var result = method.Invoke(null, arguments);
+
+                    Console.WriteLine($"Result: {result}");
```
Keboo: Do we care about the case where `result` is null? Such as the case where the method returns `void`

Keboo: Should this also be done in the above case for STA threads?

src/Shared/Program.cs
```diff
@@ -47,23 +51,35 @@ public static void Main(string[] args)
...
-                        arguments = new object[] { stringArguments.ToArray() };
+                        arguments = stringArguments.ToArray();
                     }
                 }
                 if (method.GetCustomAttributes(typeof(STAThreadAttribute), false).Any())
                 {
-                    Thread thread = new Thread(() => method.Invoke(null, arguments));
+                    object result = "";
+                    
+                    Thread thread = new Thread(() => result = method.Invoke(null, arguments));
```
Keboo:
```suggestion
                    Thread thread = new Thread(() => { 
                    object result = method.Invoke(null, arguments);
                    if (!(method.ReturnType == typeof(void)))
                    {
                        Console.WriteLine($"Result: {result}");
                    }
                    
                    });
```

PR #66, opened 2019-07-23
Add tests chapter01

### LINKED ISSUE #65: Tests Missing in Chapter01

COsborn2:
Closes #65. 

Listing 01.11A returns no output and remains "untested".

### DIFF
--- a/src/Shared/Program.cs
+++ b/src/Shared/Program.cs
@@ -39,7 +39,11 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Shared
             {
                 listing = ParseListingName(listing);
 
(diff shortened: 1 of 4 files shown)


src/Shared/Program.cs
```diff
@@ -39,7 +41,11 @@ public static void Main(string[] args)
             {
                 listing = ParseListingName(listing);
 
-                Type target = assembly.GetTypes().First(type => type.FullName.Contains(listing));
+                Type target = assembly.GetTypes().First(type =>
+                {
+                    Regex reg = new Regex($"{listing}\\.+");
```
Keboo: Same comment as the other PR. This regex should be declared outside of the First call

PR #67, opened 2019-07-23
Add tests chapter01 v8

COsborn2:
Listing 01.11A returns no output and remains "untested".

### DIFF
--- a/src/Shared/Program.cs
+++ b/src/Shared/Program.cs
@@ -42,7 +42,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Shared
             {
                 input = ParseListingName(input);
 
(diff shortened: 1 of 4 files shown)


src/Shared/Program.cs
```diff
@@ -40,7 +42,11 @@ public static void Main(string[] args)
             {
                 input = ParseListingName(input);
 
-                Type? target = assembly.GetTypes().FirstOrDefault(type => type.FullName.Contains(input));
+                Type? target = assembly.GetTypes().FirstOrDefault(type =>
+                {
+                    Regex reg = new Regex($"{input}\\.+");
```
Keboo: This will create the regex for each Type declared in the assembly. It is better to move it outside of the FirstOrDefault so it is only created once.

PR #85, opened 2019-08-20
Resolve net core shared nullability

COsborn2:


src/Shared/NetCore.cs
```diff
@@ -1,15 +1,16 @@
...
-}
+﻿using System;
+using System.Reflection;
+
+public class NetCore
+{
+    public static string GetNetCoreVersion()
+    {
+        Assembly assembly = typeof(System.Runtime.GCSettings).GetTypeInfo().Assembly;
+        var assemblyCodeBase = assembly.CodeBase ?? throw new ArgumentNullException("assemblyCodeBase");
```
Keboo, line 9: I don't think this is the correct exception type since "assemblyCodeBase" is not an argument. I would suggest using an `InvalidOperationException`.

src/Shared/NetCore.cs
```diff
@@ -1,15 +1,16 @@
...
+        string[] assemblyPath = assemblyCodeBase.Split(new[] { '/', '\\' }, StringSplitOptions.RemoveEmptyEntries);
+        int netCoreAppIndex = Array.IndexOf(assemblyPath, "Microsoft.NETCore.App");
+        if (netCoreAppIndex > 0 && netCoreAppIndex < assemblyPath.Length - 2)
```
Keboo, line 12: I realize this is the same as it was. But this code fails on *nix systems if Microsoft.NETCore.App is installed at the root. 
"/Microsoft.NETCore.App"

PR #93, opened 2019-09-01
Add tests chapter08 v8

COsborn2:


src/Chapter08.Tests/Listing08.03.ImplementingAnInterface.cs
```diff
@@ -0,0 +1,63 @@
...
+        public void CompareTo_ObjLargerThanThis_ReturnNegative1()
+        {
+            Contact firstContact = new Contact("Inigo Montoya");
+            Contact secondContact = new Contact("Zero Zero");
+            
+            Assert.AreEqual(-1, firstContact.CompareTo(secondContact));
+        }
+
+        [TestMethod]
+        public void CompareTo_ObjLargerThanThis_Return1()
```
Keboo: This test and the one above appear to have the same condition name. One should be renamed

src/Chapter08/Listing08.03.ImplementingAnInterface.cs
```diff
@@ -71,6 +72,33 @@ string[] IListable.ColumnValues
         }
         #endregion
 
+        protected new string Name
```
Keboo: Why use `new` instead of `override`?

src/Chapter08/Listing08.03.ImplementingAnInterface.cs
```diff
@@ -71,6 +72,33 @@ string[] IListable.ColumnValues
...
+            get
+            {
+                return FirstName + " " + LastName;
+            }
+            set
+            {
+                // Split the assigned value into 
+                // first and last names
+                string[] names;
+                names = value.Split(new char[] { ' ' });
```
Keboo, line 86: Did the original not combine the declaration and assignment?

COsborn2: It did not. I lifted the Name property from a previous listing. Other properties were not included in the listing so I didn't have to make a book update. So I can change the property to an override above as well.

PR #99, opened 2019-09-24
Add gitattributes to enforce Unix EOL

COsborn2:


.gitattributes
```diff
@@ -0,0 +1,14 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.c text
+*.h text
```
Keboo: Should this be *.cs?

.gitattributes
```diff
@@ -0,0 +1,14 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.c text
```
Keboo: We should probably include csproj to this too

Keboo (approved):

PR #109, opened 2019-10-07
Cryptographer build fix

COsborn2:


src/Chapter19/Cryptographer.cs
```diff
@@ -27,9 +27,9 @@ public Cryptographer()
         }
         #endregion CONSTRUCTORS
 
-        public byte[] Encrypt(string text)
+        public string Encrypt(string text)
         {
-            return EncryptAsync(text).Result;
+            return Encoding.UTF8.GetString(EncryptAsync(text).Result);
```
Keboo, line 32: Why convert this to a string just to convert it back to bytes in the usage?

PR #110, opened 2019-10-11
Refactored/reimplemented Chapter 19 cryptography stuff

MarkMichaelis:


src/Chapter19/Cryptographer.cs
```diff
@@ -111,7 +111,7 @@ public async Task DecryptAsync(byte[] encryptedData, Stream outputStream)
             await DecryptAsync(encryptedData, CryptoAlgorithm.CreateDecryptor(), outputStream);
         }
 
-        public static async Task<string> DecryptAsync(byte[] encryptedData, byte[] key, byte[] iV)
+        static public async Task<string> DecryptAsync(byte[] encryptedData, byte[] key, byte[] iV)
```
Keboo, line 114: Is there a reason for putting static before public here? It is [more common to have the access modifier first](https://stackoverflow.com/questions/191929/is-there-a-convention-to-the-order-of-modifiers-in-c)

src/Chapter19/Cryptographer.cs
```diff
@@ -27,9 +27,9 @@ public Cryptographer()
         }
         #endregion CONSTRUCTORS
 
-        public string Encrypt(string text)
+        public byte[] Encrypt(string text)
         {
-            return Encoding.UTF8.GetString(EncryptAsync(text).Result);
+            return EncryptAsync(text).Result;
```
Keboo, line 32: Is this code shown in the book? This can cause a deadlock if this code is invoked from a program with a synchronization context. If we have to force sync over async, this is a better pattern as it forces the execution onto a background thread.

```suggestion
            return Task.Run(() => EncryptAsync(text)).Result;
```

Keboo (approved):

src/Chapter19/Cryptographer.cs
MarkMichaelis: Hmmm.... Great question... and something we should include in IntelliTect analysis.
I prefer static first because it stands from the normal (which is non-static). In other words, by putting static first, it is more noticeable than when it is buried after public.  That said, I have seen code analysis rules suggesting to put public first - making this another issue similar to private field naming.

src/Chapter19/Cryptographer.cs
MarkMichaelis: It's a console program so there is no synchronization context, however, you make a good point.  I will document this as something to review when I get to that chapter.

src/Chapter19/Cryptographer.cs
Keboo: I would say log something in the [Coding Standards](https://github.com/IntelliTect/CodingStandards) and label it with "proposal". I think before we change from what the default editorconfig and other static tools suggest we should have a good reason though.

MarkMichaelis: Done: https://github.com/IntelliTect/CodingStandards/issues/42

PR #123, opened 2020-03-31
Remove VT100 chars before Assert

COsborn2:


Keboo:
This works but is there a reason we are parsing the string twice to do the removals? Why not do the char 27 removal with the regex too so you only process the string once?

Keboo (approved):

PR #125, opened 2020-04-18
Addressed extra Read() when task was completed

MarkMichaelis:
Added an `ConsoleReadAsync()` method but would appreciate a second pair of eyes on the implementation.  The purpose of the method is to provide a way to simulate cancelling a console.Read().

### DIFF
--- a/src/Chapter21/Listing21.10.CancelingParallelLoop.cs
+++ b/src/Chapter21/Listing21.10.CancelingParallelLoop.cs
@@ -50,7 +50,6 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
                 // Cancel which ever task has not finished.
                 cts.Cancel();
                 await task;
-                await Task.FromCanceled(cts.Token);
 
                 Console.ForegroundColor = ConsoleColor.Green;
                 Console.WriteLine("\nCompleted successfully");
@@ -68,27 +67,26 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
         private static async Task<int> ConsoleReadAsync(
             CancellationToken cancellationToken = default)
         {
-            int result = 0;
-            await Task.Run(async () =>
+            return await Task.Run(async () =>
             {
                 const int maxDelay = 1025;
-                int delay = 1;
+                int delay = 0;
                 while (!cancellationToken.IsCancellationRequested)
                 {
                     if (Console.KeyAvailable)
                     {
-                        result = Console.Read();
-                        break;
+                        return Console.Read();
                     }
                     else
                     {
                         await Task.Delay(delay,cancellationToken);
-                        if (delay < maxDelay) delay *= 2;
+                        if (delay < maxDelay) delay *= 2 + 1;
                     }
                 }
+                cancellationToken.ThrowIfCancellationRequested();
+                throw new InvalidOperationException(
+                    "Previous line should throw preventing this fromever executing");
             }, cancellationToken);
-            cancellationToken.ThrowIfCancellationRequested();
-            return result;
         }
 
         private static string Encrypt(string item)
--- a/src/Chapter21/Listing21.10.CancelingParallelLoop.cs
+++ b/src/Chapter21/Listing21.10.CancelingParallelLoop.cs
@@ -33,7 +33,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
             ConsoleColor originalColor = Console.ForegroundColor;
             List<string> data = Utility.GetData(100000).ToList();
 
-            CancellationTokenSource cts =
+            using CancellationTokenSource cts =
                 new CancellationTokenSource();
 
             Task task = Task.Run(() =>
@@ -41,7 +41,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
                 data = ParallelEncrypt(data, cts.Token);
             }, cts.Token);
 
-            Console.WriteLine("Push ENTER to Exit.");
+            Console.WriteLine("Push any key to Exit.");
             Task<int> cancelTask = ConsoleReadAsync(cts.Token);
 
             try
@@ -50,6 +50,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
                 // Cancel which ever task has not finished.
                 cts.Cancel();
                 await task;
+                await Task.FromCanceled(cts.Token);
 
                 Console.ForegroundColor = ConsoleColor.Green;
                 Console.WriteLine("\nCompleted successfully");
@@ -70,8 +71,8 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
             int result = 0;
             await Task.Run(async () =>
             {
-                int maxDelay = 1025;
-                int delay = 0;
+                const int maxDelay = 1025;
+                int delay = 1;
                 while (!cancellationToken.IsCancellationRequested)
                 {
                     if (Console.KeyAvailable)
@@ -81,7 +82,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
                     }
                     else
                     {
-                        await Task.Delay(delay);
+                        await Task.Delay(delay,cancellationToken);
                         if (delay < maxDelay) delay *= 2;
                     }
                 }
--- a/src/Chapter21/Listing21.10.CancelingParallelLoop.cs
+++ b/src/Chapter21/Listing21.10.CancelingParallelLoop.cs
@@ -50,7 +50,6 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter21.Listing21_10
                 // Cancel which ever task has not finished.
                 cts.Cancel();
                 await task;
-                await cancelTask;
 
                 Console.ForegroundColor = ConsoleColor.Green;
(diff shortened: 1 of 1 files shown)


src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
+                Console.WriteLine("\nCompleted successfully");
+
+            }
+            catch(OperationCanceledException taskCanceledException)
+            {
+                Console.ForegroundColor = ConsoleColor.Red;
+                Console.WriteLine(
+                    $"\nCancelled: { taskCanceledException.Message }");
+            }
+            Console.ForegroundColor = originalColor;
```
Keboo, line 64: Do you care to put this in a finally block to make sure it properly gets set back?

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
-                        innerException =>
-                            innerException.GetType() ==
-                            typeof(TaskCanceledException));
-                    if(taskCanceledException != null){
-                        Console.WriteLine($@"Cancelled: { 
-                            taskCanceledException.Message }");
+                    if (Console.KeyAvailable)
+                    {
+                        result = Console.Read();
+                        break;
```
Keboo: The output above says to press enter to exit, but this will break on all key presses.

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
+            Console.ForegroundColor = originalColor;
+        }
+
+        private static async Task<int> ConsoleReadAsync(
+            CancellationToken cancellationToken = default)
+        {
+            int result = 0;
+            await Task.Run(async () =>
+            {
+                int maxDelay = 1025;
```
Keboo:
```suggestion
                const int maxDelay = 1025;
```

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
                     }
                     else
                     {
-                        string message =
-                            string.Join(
-                                Environment.NewLine, exception.Flatten().InnerExceptions.Select(
-                            eachException => $"\t{ eachException.Message }").Distinct());
-                        Console.WriteLine($"ERROR(s): { Environment.NewLine }{ message }");
+                        await Task.Delay(delay);
+                        if (delay < maxDelay) delay *= 2;
```
Keboo: `delay` starts at 0 so `delay *= 2` is always going to be zero.

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
+            catch(OperationCanceledException taskCanceledException)
+            {
+                Console.ForegroundColor = ConsoleColor.Red;
+                Console.WriteLine(
+                    $"\nCancelled: { taskCanceledException.Message }");
+            }
+            Console.ForegroundColor = originalColor;
+        }
+
+        private static async Task<int> ConsoleReadAsync(
```
Keboo, line 67: This returns an `int` but the return value does not appear to get used.

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
+                        break;
                     }
                     else
                     {
-                        string message =
-                            string.Join(
-                                Environment.NewLine, exception.Flatten().InnerExceptions.Select(
-                            eachException => $"\t{ eachException.Message }").Distinct());
-                        Console.WriteLine($"ERROR(s): { Environment.NewLine }{ message }");
+                        await Task.Delay(delay);
```
Keboo: Pass the cancellation token to `Task.Delay` so it can quit early if it is cancelled.

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
+                Console.WriteLine(
+                    $"\nCancelled: { taskCanceledException.Message }");
+            }
+            Console.ForegroundColor = originalColor;
+        }
+
+        private static async Task<int> ConsoleReadAsync(
+            CancellationToken cancellationToken = default)
+        {
+            int result = 0;
```
Keboo: If there a reason for capture this variable rather than having the delegate inside of the `Task.Run` return the int, and this method simply return the Task.Run?

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
             CancellationTokenSource cts =
```
Keboo: CancellationTokenSource is IDisposable.
```suggestion
            using CancellationTokenSource cts = new CancellationTokenSource();
```

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -35,52 +36,58 @@ public static async Task Main()
...
 
-            if (!task.IsCompleted)
+            try
             {
+                Task.WaitAny(task, cancelTask);
+                // Cancel which ever task has not finished.
                 cts.Cancel();
-                try { task.Wait(); }
-                catch (AggregateException exception)
+                await task;
```
Keboo, line 52: If the task is cancelled you cannot await it. Consider this example:
```
using var cts = new CancellationTokenSource();
var token = cts.Token;
cts.Cancel();
var t = Task.FromCanceled(token);
await t.ConfigureAwait(false);
```

Keboo (changes requested):

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
MarkMichaelis: @Keboo: Are you suggesting this because I didn't await the `cancelTask`?
I do await task intentionally, knowing that if it is cancelled it will throw an exception.
It isn't clear what the `Task.FromCanceled(token)` gives me in this case.

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
MarkMichaelis: Indeed... I was trying to write a generic function.

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
Keboo: Because you call cts.Cancel just above this, doesn't that always mean this will throw?

MarkMichaelis: Thanks for the discussion @Keboo .

Firstly, the code has changed significantly such that the comment may no longer apply.  Here's the updated code:
```
            try
            {
                Task.WaitAny(task, cancelTask);
                // Cancel whichever task has not finished.
                cts.Cancel();
                await task;

                Console.ForegroundColor = ConsoleColor.Green;
                Console.WriteLine("\nCompleted successfully");

            }
            catch (OperationCanceledException taskCanceledException)
            {
                Console.ForegroundColor = ConsoleColor.Red;
                Console.WriteLine(
                    $"\nCancelled: { taskCanceledException.Message }");
            }
            finally
            {
                Console.ForegroundColor = originalColor;
            }
```

Awaiting a cancelled task will indeed throw, and I count on this in order to display the cancelled message when the user cancels the activity.

Perhaps the issue is that I don't ever await the `cancelTask`?  If I did `await cancelTask` I would  always catch and suppress the exception because if `task` runs to completion, I have to cancel the `cancelTask` rather than continue waiting for a keyboard entry. 

**Is it necessary to always await all tasks even if you know they have completed and you don't care about capturing the exception?**

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
```diff
@@ -32,55 +33,60 @@ public static async Task Main()
...
+                        return Console.Read();
                     }
                     else
                     {
-                        string message =
-                            string.Join(
-                                Environment.NewLine, exception.Flatten().InnerExceptions.Select(
-                            eachException => $"\t{ eachException.Message }").Distinct());
-                        Console.WriteLine($"ERROR(s): { Environment.NewLine }{ message }");
+                        await Task.Delay(delay,cancellationToken);
```
Keboo, line 82: nit: spacing.
```c#
await Task.Delay(delay,cancellationToken);
```

src/Chapter21/Listing21.10.CancelingParallelLoop.cs
Keboo: It just seems like you are then using exceptions for control flow (something discouraged in earlier chapters). If the goal is to simply display a message, wouldn't it be better to simply check the state of the token, and display the message? 
Manually setting up a continuation on the `task` would also seem like a preferable options to the exception handler.

MarkMichaelis:
@Keboo, that is a great point.  In this case the purpose of using exceptions is to demonstrate how control flows during cancellation.  To not do so would be to miss the point that invoking `cts.Cancel()` throws and exception and how to catch it.
Thanks again so much for the feedback. It is hugely valuable.  Each an every comment you make has caused me to think/review, make a change or a clarification.  Thanks so much!!!

PR #133, opened 2021-01-07
Think there might be a typo on pg 12 

SalimGangji:
Think there is a typo on pg12, the text should say `-p:PublishSingleFile=true` instead of `/p:PublishSingleFile=true`

### DIFF
--- a/Errata.md
+++ b/Errata.md
@@ -11,6 +11,7 @@ The following corrections will be made in the second printing. (To determine whi
 
 Found by            | Chapter     | Page        | Correction                                                                                                                                                       
 ------------------- | ----------- | ----------- | -----------------------------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


BenjaminMichaelis:
The change suggested from a /p: to a -p:
Both work, however the -p: is an improvement as it is more universal, not working only in windows
The change suggested in the errata will be updated in the manuscript

SalimGangji:
Awesome, thank you Benjamin 👍

Errata.md
```diff
@@ -11,6 +11,7 @@ The following corrections will be made in the second printing. (To determine whi
 
 Found by            | Chapter     | Page        | Correction                                                                                                                                                       
 ------------------- | ----------- | ----------- | -----------------------------------------------------------------------------------------------------------------------------------------------------------------
+Salim Gangji        | 1           | 12          | First paragraph: For a single executable, append <s>`/p:PublishSingleFile=true`</s> `-p:PublishSingleFile=true` to the command
```
BenjaminMichaelis, line 14: The change suggested from a /p: to a -p:
Both work, however the -p: is an improvement as it is more universal, not working only in windows as the /p: does

PR #134, opened 2021-01-10
Potential typo in Chapter 2 pg51 output 2.3

SalimGangji:


BenjaminMichaelis:
Change suggested for output 2.3 is correct. 
After checking the expected answer listed in the paragraph after the output, running the code itself in Listing 2.3, and checking the tests, the output should be 1.618033988749895 as is being suggested in this change. 
Change will be applied to the manuscript.

PR #142, opened 2021-01-19
Updated Listing 03.01

BenjaminMichaelis:


BenjaminMichaelis:
Old version is in print v8.0 but also doesn't have a test, so maybe changing it is beneficial?

PR #166, opened 2021-06-01
Add report generator to linux build

BenjaminMichaelis:


BenjaminMichaelis:
The workflow is not valid. .github/workflows/Branch v9.0 - EssentialCSharp-Linux.yml (Line: 31, Col: 13): Expected format {org}/{repo}[/path]@ref. Actual 'reportgenerator@4' Input string was not in a correct format.

PR #181, opened 2021-06-09
Updated Chapters 2/3/4 with corrections

kNoAPP:
Errata: Added log of significant changes.
2.15: Added public identifiers to correct Reflection error.
2.17: Added a space in code to match book.
3.9/10/11/12/20/21/27/28: Replaced "Pascal" with "Visual Basic" to match book.
3.27: Removed whitespace for the additional example. 
4.18: Code comments for incrementing example were incorrect. They've been updated to match the book.
Program: Added "Listing" to Regex because it confused Listing.cs files with Table.cs files causing Listings 4.4/7.3/etc. to fail.

### DIFF
--- a/src/Chapter03/Listing03.27.Slicing.cs
+++ b/src/Chapter03/Listing03.27.Slicing.cs
@@ -6,7 +6,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter03.Listing03_27
         {
             string[] languages = new string[] {
                 "C#", "COBOL", "Java",
-                "C++", "TypeScript", "Visual Basic",
+                "C++", "TypeScript", "Swift",
                 "Python", "Lisp", "JavaScript"};
 
             System.Console.WriteLine($@"  0..3: {
@@ -16,7 +16,7 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter03.Listing03_27
                 string.Join(", ", languages[^3..^0]) // Python, Lisp, JavaScript
             }"); 
             System.Console.WriteLine($@" 3..^3: {
-                string.Join(", ", languages[3..^3]) // C++, TypeScript, Visual Basic
+                string.Join(", ", languages[3..^3]) // C++, TypeScript, Swift
             }");
             System.Console.WriteLine($@"  ..^6: {
                 string.Join(", ", languages[..^6])  // C#, COBOL, Java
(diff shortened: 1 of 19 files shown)


src/Chapter03/Listing03.27.Slicing.cs
```diff
@@ -6,7 +6,7 @@ public static void Main()
         {
             string[] languages = new string[] {
                 "C#", "COBOL", "Java",
-                "C++", "TypeScript", "Pascal",
+                "C++", "TypeScript", "Visual Basic",
```
BenjaminMichaelis: Would you be willing to update the test [Chapter3.27 Test](https://github.com/IntelliTect/EssentialCSharp/blob/various-additional-fixes/src/Chapter03.Tests/Listing03.27.Slicing.cs) or allow me to correct it?

src/Chapter03/Listing03.28.AdditionalArrayMethods.cs
```diff
@@ -6,7 +6,7 @@ public static void Main()
         {
             string[] languages = new string[]{
                 "C#", "COBOL", "Java",
-                "C++", "TypeScript", "Pascal",
+                "C++", "TypeScript", "Visual Basic",
```
BenjaminMichaelis: Would you be willing to update the test [Chapter3.38 Test](https://github.com/IntelliTect/EssentialCSharp/blob/various-additional-fixes/src/Chapter03.Tests/Listing03.38.ReversingAString.cs) which also needs the file to be renamed to "Listing03.28.ReversingAString.cs" or allow me to correct it?

src/Chapter04/Listing04.02.Negative.cs
```diff
@@ -5,7 +5,7 @@ public class Program
         public static void Main()
         {
             //National Debt to the Penny
-            decimal debt = -18125876697430.99M;
+            decimal debt = -28382607800141.79M;
```
BenjaminMichaelis: I don't see this as being the correct number in the book for this listing? It looks like to me that the correct number should be "-26457079712930.80M"

BenjaminMichaelis (changes requested):
Some good changes here! Thank you! It looks like some of the tests need to be updated is the most pressing problem I see currently, as 3 tests are failing, but one looks like the content might of still be incorrect even after a change (4.02, I left comments on these). If you would be willing to update these tests that would be awesome, otherwise I don't mind making the changes needed and pushing them to your branch if you would like.

BenjaminMichaelis:
@kNoAPP I just wanted to follow up and see if you wanted to update this with the changes or if I should start work on it otherwise! 
Thanks!

kNoAPP:
You can start work on it! Sorry for the super late reply.

PR #207, opened 2021-12-27
Create Code-Analysis.yml

### LINKED ISSUE #17: Put back code analysis
It seems that in the conversion we lost code analysis.  It would be great to get this back up and running but it is not the highest priority by any means.

BenjaminMichaelis:


BenjaminMichaelis:
build-breaking should be made to true (build-breaking: true) once all code-analysis problems are fixed

PR #218, opened 2022-01-02
Fix warnings

### LINKED ISSUE #50: Remove #nullable from Program.cs
Once conversion to C# 8 nullable is complete, remove the #nullable declarative from Program.cs

### LINKED ISSUE #206: Test 10.13
Even though powershell returns a 0 exit code, might not be running geocordinates program due to pathing issue on mac/Linux. Need to look in further to actual powershell output.

### LINKED ISSUE #247: Handle Ctrl z to quit
On Tic Tac Toe in chapter 4

### LINKED ISSUE #249: Revisit parse management for null or empty on 5.22 and related listings
lines 22, and other listings like it (`age = int.Parse(ageText);`

BenjaminMichaelis:
Fix all warnings except the 7 expected warnings in the source code. 
Also fixes #206 and fixes #50

Expected warnings:

- Warning CS1030 in 4.58
- Warning CS1030 in 4.61
- Warning CA1507 in 6.20
- Warning CS0108 in 7.11
- Warning CA2013 in 10.04
- Warning CA2013 in 10.04
- Warning CS0612 in 18.25


src/Chapter04/Listing04.44.DoWhileLoop.cs
```diff
@@ -16,7 +16,8 @@ public static void Main()
                 // Request a move from the current player
                 System.Console.Write(
                     $"\nPlayer {currentPlayer}: Enter move:");
-                string input = System.Console.ReadLine();
+                // TODO: Update listing in Manuscript
+                string input = System.Console.ReadLine() ?? string.Empty;
```
Keboo: For this one I think it might be cleaner to simply declare `input` and nullable
```suggestion
                string? input = System.Console.ReadLine();
```

src/Chapter04/Listing04.45.ForLoop.cs
```diff
@@ -11,7 +11,9 @@ public static void Main()
             System.Console.Write("Enter an integer: ");
             // Use long.Parse() to support negative numbers
             // Assumes unchecked assignment to ulong
-            value = (ulong)long.Parse(System.Console.ReadLine());
+            // TODO: Update listing in Manuscript
+            // If ReadLine returns null, use "42" as default input
+            value = (ulong)long.Parse(System.Console.ReadLine() ?? "42");
```
Keboo: Since the sample output in the book for this one shows "-42" I would propose we use it for the default.
```suggestion
            value = (ulong)long.Parse(System.Console.ReadLine() ?? "-42");
```

src/Chapter04/Listing04.53.ReplacingContinueWithIf.cs
```diff
@@ -8,7 +8,8 @@ public static void Main()
             bool insideDomain = false;
 
             System.Console.WriteLine("Enter an email address: ");
-            email = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
```
Keboo: This one doesn't need to be updated as this line is not included in the displayed portion of the listing.

src/Chapter04/TicTacToe.cs
```diff
@@ -59,7 +59,8 @@ public class TicTacToeGame      // Declares the TicTacToeGame class.
             {
                 // Request a move from the current player.
                 System.Console.Write($"\nPlayer {currentPlayer} - Enter move:");
-                input = System.Console.ReadLine();
+                // TODO: Update listing in Manuscript
+                input = System.Console.ReadLine() ?? string.Empty;
```
Keboo: For this one I would make `input` nullable, and update the signature of `ValidateAndMove`

src/Chapter05/Listing05.02.SimpleMethodCall.cs
```diff
@@ -10,10 +10,12 @@ public static void Main()
...
 
             System.Console.Write("Enter your first name: ");
-            firstName = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            firstName = System.Console.ReadLine() ?? string.Empty;
 
             System.Console.Write("Enter your last name: ");
-            lastName = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            lastName = System.Console.ReadLine() ?? string.Empty;
```
Keboo: Because the names here are just going into an interpolated string, I would propose just declaring `firstName` and `lastName` as nullable.

src/Chapter05/Listing05.12.PassingCommandLineArgumentsToMain.cs
```diff
@@ -1,5 +1,6 @@
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_12
 {
+    // TODO: Update listing in Manuscript
```
Keboo, line 3: I don't think there is an update that is  needed i the manuscript here.

src/Chapter05/Listing05.18.CountingLinesGivenADirectory.cs
```diff
@@ -1,7 +1,6 @@
-#nullable enable
-
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_18
 {
+    // TODO: Update listing in Manuscript
```
Keboo, line 3: It doesn't looks like there are any manuscript updates needed here.

src/Chapter05/Listing05.22.ConvertingAStringToAnInt.cs
```diff
@@ -13,10 +13,12 @@ public static void Main()
...
 
             Console.Write("Enter your first name: ");
-            firstName = Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            firstName = Console.ReadLine() ?? string.Empty;
 
             Console.Write("Enter your age: ");
-            ageText = Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            ageText = Console.ReadLine() ?? string.Empty;
```
Keboo: I would make `firstName` and `ageText` nullable. Also worth noting that both the null and empty string cases will throw on the `int.Parse(ageText)`

src/Chapter05/Listing05.23.CatchingAnException.cs
```diff
@@ -12,10 +12,12 @@ public static int Main(string[] args)
...
 
             Console.Write("Enter your first name: ");
-            firstName = Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            firstName = Console.ReadLine() ?? string.Empty;
 
             Console.Write("Enter your age: ");
-            ageText = Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            ageText = Console.ReadLine() ?? string.Empty;
```
Keboo: Same here. I would make `firstName` and `ageText` nullable

src/Chapter05/Listing05.24.FinallyWithoutCatch.cs
```diff
@@ -12,10 +12,12 @@ public static int Main()
...
 
             Console.Write("Enter your first name: ");
-            firstName = Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            firstName = Console.ReadLine() ?? string.Empty;
 
             Console.Write("Enter your age: ");
-            ageText = Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            ageText = Console.ReadLine() ?? string.Empty;
```
Keboo: Same here. I would make `firstName` and `ageText` nullable

src/Chapter05/Listing05.25.GeneralCatchBlocks.cs
```diff
@@ -12,10 +12,12 @@ public static int Main()
...
 
             System.Console.Write("Enter your first name: ");
-            firstName = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            firstName = System.Console.ReadLine() ?? string.Empty;
 
             System.Console.Write("Enter your age: ");
-            ageText = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            ageText = System.Console.ReadLine() ?? string.Empty;
```
Keboo: Same here. I would make `firstName` and `ageText` nullable

src/Chapter05/Listing05.28.TryCatch.cs
```diff
@@ -8,10 +8,12 @@ public static void Main()
...
 
             System.Console.Write("Enter your first name: ");
-            firstName = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            firstName = System.Console.ReadLine() ?? string.Empty;
 
             System.Console.Write("Enter your age: ");
-            ageText = System.Console.ReadLine();
+            // TODO: Update listing in Manuscript
+            ageText = System.Console.ReadLine() ?? string.Empty;
```
Keboo: Same here. I would make `firstName` and `ageText` nullable

src/Chapter07/Chapter07.csproj
```diff
@@ -3,16 +3,21 @@
...
+  <ItemGroup>
+    <Compile Remove="Table07.03.TypePatternMatchingWithIsOperator.cs" />
+  </ItemGroup>
   <ItemGroup>
     <Compile Include="..\Shared\Program.cs">
       <Link>Program.cs</Link>
     </Compile>
   </ItemGroup>
+  <ItemGroup>
+    <None Include="Table07.03.TypePatternMatchingWithIsOperator.cs" />
```
Keboo, line 19: I would combine this into the `ItemGroup` above

src/Chapter07/Table07.03.TypePatternMatchingWithIsOperator.cs
```diff
@@ -1,10 +1,45 @@
...
+
+            // ...
+        else (GetObjectById(id) is var result)
+        {
+            // ...
+        }
+
+
+
+    // Was existent in the source code already
```
Keboo:
```suggestion
    // Was existing in the source code already
```

src/Chapter10/Listing10.05.OverridingEquals.cs
```diff
@@ -60,22 +59,20 @@ public bool Equals(Coordinate obj)
         {
             // STEP 1: Check for null if a reference type
             // (e.g., a reference type)
-            // if (obj == null)
+            // if (ReferenceEquals(obj, null)
```
Keboo: I realize this matches the manuscript, but IMO it would be better if this suggested using pattern matching for the null check to avoid boxing. @MarkMichaelis 
```suggestion
            // if (obj is null)
```

src/Chapter10/Listing10.23.RegisteringAFinalizerWithProcessExit.cs
```diff
@@ -93,12 +94,14 @@ public SampleUnmanagedResource()
             WriteLine("Exiting...");
         }
 
-        #region IDisposable Members
         public void Dispose()
         {
             Dispose(true);
+            
+            // Unregister from the finalization queue.
```
Keboo: I realize this matches the original comment, but I think I would change this wording since the idea of the finalization queue is an implementation detail that is only mostly true (since different runtimes may change this). Also, at this point in the code the object has not been add to it, instead it is preventing itself from getting added.

Perhaps something more like this (stollen from the MS docs):
```suggestion
            // Request that the finalizer not be called for this object.
```

src/Chapter13/Listing13.08.UsingMethodNameAsArgument.cs
```diff
@@ -42,12 +43,18 @@ public static bool GreaterThan(int first, int second)
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer: ");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.11.UsingADifferentFuncCompatibleMethod.cs
```diff
@@ -44,12 +45,18 @@ public static bool GreaterThan(int first, int second)
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer: ");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.12.CreatingDelegateWithStatementLambda.cs
```diff
@@ -46,12 +47,18 @@ public static bool GreaterThan(int first, int second)
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer: ");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.13.OmittingParameterTypesFromStatementLambdas.cs
```diff
@@ -8,12 +9,18 @@ public class Program
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer:");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.15.ParameterlessStatementLambdas.cs
```diff
@@ -9,14 +10,15 @@ public static void Main()
...
                 {
-                    string input;
+                    string? input;
                     do
                     {
                         input = Console.ReadLine();
                     }
-                    while(input.Trim().Length == 0);
-                    return input;
+                    while(!string.IsNullOrEmpty(input));
```
Keboo: The previous code was using `.Trim()` which would remove whitespace. So the equivalent here would be to check against null or whitespace.
```suggestion
                    while(!string.IsNullOrWhiteSpace(input));
```

src/Chapter13/Listing13.16.PassingADelegateWithAnExpressionLambda.cs
```diff
@@ -8,12 +9,18 @@ public class Program
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer:");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.17.PassingAnAnonymousMethodInCSharpTwo.cs
```diff
@@ -8,12 +9,18 @@ public class Program
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer:");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.18.UsingVarianceForDelegates.cs
```diff
@@ -17,9 +17,10 @@ public static void Main()
 
             Action<string> narrowAction = broadAction;
 
+            // TODO: Update listing in Manuscript
             // Covariance
             Func<string> narrowFunction =
-                () => Console.ReadLine();
+                () => 8654091021.ToString();
```
Keboo: If we are just going to return a hard coded string I would just use a string literal. 
```suggestion
                () => "8654091021";
```

src/Chapter13/Listing13.19.CSharpEquivalentOfCILGeneratedByTheCompilerForLambdaExpressions.cs
```diff
@@ -32,12 +33,18 @@ public class DelegateSample
...
         {
             int i;
+            string? text;
             int[] items = new int[5];
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer:");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.20.UsingAnOuterVariableInALambdaExpression.cs
```diff
@@ -8,13 +9,19 @@ public class Program
...
             int i;
+            string? text;
             int[] items = new int[5];
             int comparisonCount = 0;
 
             for(i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer:");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter13/Listing13.21.CSharpEquivalentOfCILCodeGeneratedByCompilerForOuterVariables.cs
```diff
@@ -24,11 +25,17 @@ public static void Main()
...
             locals.comparisonCount = 0;
             int[] items = new int[5];
+            string? text;
 
-            for(i = 0; i < items.Length; i++)
+            for (i = 0; i < items.Length; i++)
             {
                 Console.Write("Enter an integer:");
-                items[i] = int.Parse(Console.ReadLine());
+                text = Console.ReadLine();
```
Keboo: I would combine declaration as assignment of `text` (might even go as far to say the same should be done for `i`)

src/Chapter14/Listing14.03.ConnectingThePublisherAndSubscribers.cs
```diff
@@ -11,7 +12,7 @@ public static void Main()
             Thermostat thermostat = new Thermostat();
             Heater heater = new Heater(60);
             Cooler cooler = new Cooler(80);
-            string temperature;
+            string? temperature;
```
Keboo: I would combine declaration as assignment of `temperature`

src/Chapter14/Listing14.09.OnTemperatureChangedThrowingAnException.cs
```diff
@@ -11,7 +12,7 @@ public static void Main()
             Thermostat thermostat = new Thermostat();
             Heater heater = new Heater(60);
             Cooler cooler = new Cooler(80);
-            string temperature;
+            string? temperature;
```
Keboo: I would combine declaration as assignment of `temperature`

src/Chapter14/Listing14.10.HandlingExceptionsFromSubscribers.cs
```diff
@@ -59,7 +60,7 @@ public static void Main()
                 Thermostat thermostat = new Thermostat();
                 Heater heater = new Heater(60);
                 Cooler cooler = new Cooler(80);
-                string temperature;
+                string? temperature;
```
Keboo: I would combine declaration as assignment of `temperature`

src/Chapter14/Listing14.11.UsingTheAssignmentOperationRatherThanPlusEquals.cs
```diff
@@ -12,7 +13,7 @@ public static void Main()
             Thermostat thermostat = new Thermostat();
             Heater heater = new Heater(60);
             Cooler cooler = new Cooler(80);
-            string temperature;
+            string? temperature;
```
Keboo: I would combine declaration as assignment of `temperature`

src/Chapter16/Listing16.05.DeferredExecutionAndQueryExpressions2.cs
```diff
@@ -55,7 +55,8 @@ where keyword.Contains('*')
 
             // Retrieve the count from the cached collection
             Console.WriteLine(
-                $"7. selectionCache count={ selectionCache.Count() }");
+                // TODO: Update listing in Manuscript
```
Keboo, line 58: I don't think there is a manuscript change needed here.

src/Chapter18/18.11.SpecifyingAReturnAttribute.DescriptionAttribute.cs
```diff
@@ -4,8 +4,8 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter18.Listing18_11
 {
     public class DescriptionAttribute : Attribute
     {
-        private string description;
-		public string Description { get { return Description; } }
+        private readonly string description;
+		public string Description { get { return description; } }
```
Keboo: I realize this matches the original, but we should fix the naming of the field.
```suggestion
        private readonly string _Description;
	public string Description { get { return _Description; } }
```

src/Chapter19/Listing19.07.RegisteringForUnhandledExceptions.cs
```diff
@@ -6,12 +6,12 @@
 
     public class Program
     {
-        public static Stopwatch _Clock = new Stopwatch();
+        private static Stopwatch _Clock { get;  } = new Stopwatch();
```
Keboo: We should fix the name of the property here. 
```suggestion
        private static Stopwatch Clock { get; } = new Stopwatch();
```

src/Chapter20/Listing20.12.CustomMethodWithProgressSupport.cs
```diff
@@ -10,10 +11,10 @@ class Program
         static public Task<Process> RunProcessAsync(
             string fileName,
             string arguments = "",
+            IProgress<ProcessProgressEventArgs>? progress = null,
+                object? objectState = null,
```
Keboo: nit: whitespace
```suggestion
            object? objectState = null,
```

src/Chapter23/Listing23.20.DesignatingABlockForUnsafeCode.cs
```diff
@@ -54,7 +54,7 @@ public unsafe static int Main()
         }
     }
 
-    public class VirtualMemoryPtr : System.Runtime.InteropServices.SafeHandle
+    public class VirtualMemoryPtr : System.Runtime.InteropServices.SafeHandle, IDisposable
```
Keboo: The interface declaration is technically not needed here as the base class already implements `IDisposable`. This also means you don't need to create the full dispose pattern in this class.

src/Chapter23/Listing23.20.DesignatingABlockForUnsafeCode.cs
```diff
@@ -90,15 +90,37 @@ public override bool IsInvalid
...
+                    VirtualMemoryManager.VirtualFreeEx(ProcessHandle,
+                        AllocatedPointer, MemorySize);
+                }
                 Disposed = true;
-                GC.SuppressFinalize(this);
-                VirtualMemoryManager.VirtualFreeEx(ProcessHandle,
-                    AllocatedPointer, MemorySize);
             }
-            return true;
+        }
```
Keboo: This class is deriving from SafeHandle, so it is a bit of a special case in terms of memory clean up. Because the base class implements the IDisposable pattern, it exposes the regular Dispose(bool) method that you can override to clean up. In addition, the abstract base class exposes the ReleaseHandle method (which is also designed for cleanup). Making things just a little bit harder, if you dig into the base dispose method you will find that it ends in some extern calls that get implemented by the runtime (making it tricky to tell what is actually happening). Lot of details in these docs here: SafeHandle Class (System.Runtime.InteropServices) | Microsoft Docs. Take note of the section at the very bottom titled:  Notes to Implementers 
What is convenient here is we can assume that MS has implemented the dispose correctly, so the amount of code we need to write is significantly reduced. Let’s start by looking at the regular Dispose(bool) (I am just using the default VS generates). This will be functionally similar to what SafeHandle will do:
```C#
protected virtual void Dispose(bool disposing)
{
    if (!_disposedValue)
    {
        if (disposing)
        {
            // TODO: dispose managed state (managed objects)
        }

        // TODO: free unmanaged resources (unmanaged objects) and override finalizer
        // TODO: set large fields to null
        _disposedValue = true;
    }
}
```
The Dispose implementation inside of SafeHandle will ensure that Dispose gets appropriately called (passing the right disposing value). In the TODO section for freeing unmanaged resources is where it will call ReleaseHandle (expecting that it will do the needful to cleanup unmanaged handles). You will also notice that it does this regardless of whether the disposing flag. So for our code we only need to implement ReleaseHandle, and don’t need to do any of the regular disposable members as those are already handled for us.
So for our code here the implementation we want should look like this:
```C#
protected override bool ReleaseHandle()
{
    Disposed = true;
    VirtualMemoryManager.VirtualFreeEx(ProcessHandle, AllocatedPointer, MemorySize);
}
```

src/Shared/Program.cs
```diff
@@ -84,24 +84,24 @@ public static void Main(string[] args)
...
-                        output = InvokeMethodUsingReflection(method,arguments).GetAwaiter().GetResult();
+                        output = await InvokeMethodUsingReflectionAsync(method, arguments);
                     });
                     task.Wait();
                 }
                 else
                 {
                     // TODO: Change to use async/await.
-                    output = InvokeMethodUsingReflection(method, arguments).GetAwaiter().GetResult();
+                    output = await InvokeMethodUsingReflectionAsync(method, arguments);
```
Keboo: I think entire if-condition can be replaced with:
```C#
output = await InvokeMethodUsingReflectionAsync(method, arguments);
```

The extra Tasks don't appear to be adding anything, and the STAAttribute is getting ignored and wont be used anyway. If all of the tests are passing we should b esafe to remove it.

src/Chapter23/Listing23.20.DesignatingABlockForUnsafeCode.cs
BenjaminMichaelis: I got build errors by removing this because lower down we have this function within here that calls IDisposable so I am keeping it for now:
```
        void IDisposable.Dispose()
        {
            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
            Dispose(disposing: true);
            GC.SuppressFinalize(this);
        }
```

BenjaminMichaelis:
Just wanted to get a double check over these last like 4 commits since the review just as a final double check

src/Chapter23/Listing23.20.DesignatingABlockForUnsafeCode.cs
Keboo: Correct, I would suggest that we should remove that as well

src/Chapter20.Tests/Listing20.11.ImplementingCustomAsynchronousMethod.Tests.cs
```diff
@@ -1,10 +1,8 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
 using System.Text.RegularExpressions;
-using System.Threading.Tasks;
-using AddisonWesley.Michaelis.EssentialCSharp.Chapter20.Listing20_09;
 
-namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter20.Listing20_11.Tests
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter20.Listing20_09.Tests
```
Keboo, line 5: Was this change intentional? Is there a reason this doesn't match the file name?

src/Chapter23/Listing23.20.DesignatingABlockForUnsafeCode.cs
```diff
@@ -91,14 +91,20 @@ public override bool IsInvalid
...
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: false);
+        }
+
+        void IDisposable.Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
```
Keboo: I would propose removing these as well since the SafeHandle already defines them.

src/Shared/Program.cs
```diff
@@ -84,24 +84,10 @@ public static void Main(string[] args)
...
-                    task.Wait();
-                }
-                else
-                {
-                    // TODO: Change to use async/await.
-                    output = InvokeMethodUsingReflection(method, arguments).GetAwaiter().GetResult();
-                }
-                if(output is { })
+
+                output = await InvokeMethodUsingReflectionAsync(method, arguments);
```
Keboo: nit: Combine declaration and assignment.

```C#
string? output = await InvokeMethodUsingReflectionAsync(method, arguments);
```

Keboo (commented):
Just a couple more comments then it is looking good.

src/Chapter20.Tests/Listing20.11.ImplementingCustomAsynchronousMethod.Tests.cs
BenjaminMichaelis: The file name was going to get updated here: #246 but I can do it in this one instead just for continuity

Keboo: Ah, either is fine by me.

src/Chapter05/Listing05.12.PassingCommandLineArgumentsToMain.cs
BenjaminMichaelis: This one does need to be for lines 4-6

src/Chapter05/Listing05.18.CountingLinesGivenADirectory.cs
BenjaminMichaelis: Line 1 in the book needs to be removed.

src/Chapter05/Listing05.22.ConvertingAStringToAnInt.cs
BenjaminMichaelis: Made an issue to follow up the parse

src/Chapter10/Listing10.23.RegisteringAFinalizerWithProcessExit.cs
BenjaminMichaelis: Sounds great to me!

src/Chapter16/Listing16.05.DeferredExecutionAndQueryExpressions2.cs
BenjaminMichaelis: This needs to be switched from a function to a property in the book

src/Chapter10/Listing10.05.OverridingEquals.cs
BenjaminMichaelis: Needs to be done throughout the book, but changed it here

Keboo (approved):

BenjaminMichaelis:
@Keboo Should have them all resolved now

PR #252, opened 2022-04-01
Chapter16.01 listing update

### LINKED ISSUE #251: Listing 16.01 Update for manuscript parser
The CSharp class is shown along with the code for Listing 16.1 in the manuscript. Chapter16.Listing16.CSharp needs to be moved into the same file as Chapter16.Listing16_01 in order to achieve this same effect for future iterations of the manuscript.

image

grantdwoods:
Updates to listing 16.1 so parser output will match previous pdf version

Fixes #251


BenjaminMichaelis (approved):
Looks great!

PR #272, opened 2022-05-10
Add regions to chapter 11

BenjaminMichaelis:
Contributes to Intellitect/EssentialCSharpManuscript#116


BenjaminMichaelis:
adding @MarkMichaelis to check if 11.1 is as you would like

Keboo (approved):

PR #283, opened 2022-06-30
Add missing listings

BenjaminMichaelis:
## Summary

Contributes to IntelliTect/EssentialCSharpManuscript#284 by adding in missing listings and cleaning up the listings enough for the checks


src/Chapter18/Listing18.11.SpecifyingAReturnAttribute.cs
```diff
@@ -1,13 +1,27 @@
+using System;
```
Keboo, line 1: Should this one also have the `#region INCLUDE` tags?

src/Chapter23/Listing23.06.Win32ErrorHandling.cs
```diff
@@ -0,0 +1,88 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter23.Listing23_06
+{
+    #region INCLUDE
+    public class VirtualMemoryManager
+    {
+        [DllImport("kernel32.dll SetLastError = true")]
```
Keboo:
```suggestion
        [DllImport("kernel32.dll", SetLastError = true)]
```

src/Chapter23/Listing23.07.ManagedResourcesUsingSafeHandle.cs
```diff
@@ -0,0 +1,56 @@
...
+            _AllocatedPointer =
+                VirtualMemoryManager.AllocExecutionBlock(
+                memorySize, ProcessHandle);
+            _Disposed = false;
+        }
+
+        public readonly IntPtr _AllocatedPointer;
+        readonly IntPtr _ProcessHandle;
+        readonly IntPtr _MemorySize;
+        bool _Disposed;
```
Keboo: Should these have access modifiers?

Keboo (approved):

src/Chapter18/Listing18.11.SpecifyingAReturnAttribute.cs
BenjaminMichaelis: Yes, but I was just going to let Hanna and Emma add those when they get to that chapter so I don't throw them off since they seem to do a better job of checking for those than myself.

src/Chapter23/Listing23.07.ManagedResourcesUsingSafeHandle.cs
BenjaminMichaelis:
```suggestion
        public readonly IntPtr _AllocatedPointer;
        private readonly IntPtr _ProcessHandle;
        private readonly IntPtr _MemorySize;
        private bool _Disposed;
```

PR #297, opened 2022-07-12
Chapter23 regions

HannaMichaelis:


src/Chapter23/Chapter23.csproj
```diff
@@ -5,6 +5,8 @@
   </PropertyGroup>
   <Import Project="..\Chapter.props" />
   <ItemGroup>
+    <Compile Remove="Listing23 - Copy.02.TheVirtualAllocExAPI.cs" />
```
Keboo: We should be able to remove all changes to this file. You can see them if you double click on the Chapter 23 in your Solution Explorer.

Keboo (changes requested):

BenjaminMichaelis (approved):

PR #316, opened 2022-07-26
Fix Build Failure

### LINKED ISSUE #312: Investigate why build failure is occurring only on mac
Investigate sudden failures after previously they were passing as seen here (there are two runs on the same commit):
https://github.com/IntelliTect/EssentialCSharp/runs/7528181670

BenjaminMichaelis:
fixes #312

### DIFF
--- a/.github/workflows/Branch v9.0.yml
+++ b/.github/workflows/Branch v9.0.yml
@@ -23,7 +23,7 @@ jobs:
         with:
           dotnet-version: | 
             3.1.x
-            5.0.x
+            5.x
       - name: Get Build Information
         run: |
           pwd
--- a/.github/workflows/Branch v9.0.yml
+++ b/.github/workflows/Branch v9.0.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Setup dotnet SDK
-        uses: actions/setup-dotnet@v1
+        uses: actions/setup-dotnet@v2
         with:
           dotnet-version: | 
             3.1.x
--- a/.github/workflows/Branch v9.0.yml
+++ b/.github/workflows/Branch v9.0.yml
@@ -24,6 +24,7 @@ jobs:
           dotnet-version: | 
             3.1.x
             5.x
+            6.x
       - name: Get Build Information
         run: |
           pwd
--- a/.github/workflows/Branch v9.0.yml
+++ b/.github/workflows/Branch v9.0.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup dotnet SDK
         uses: actions/setup-dotnet@v2
         with:
--- a/global.json
+++ b/global.json
@@ -1,0 +1,7 @@
+{
+  "sdk": {
+    "version": "6.0.0",
+    "rollForward": "latestMinor",
+    "allowPrerelease": false
+  }
+}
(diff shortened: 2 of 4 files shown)


BenjaminMichaelis:
Updating checkout to v3 and setup dotnet to v2

BenjaminMichaelis:
After adding the props file in, it caused the ubuntu and windows builds to fail which is interesting since that means adding in a props file fails on dotnet 5 (the build is here: https://github.com/IntelliTect/EssentialCSharp/runs/7532253715?check_suite_focus=true) but after adding in a global.json and installing the dotnet 6 sdk onto the machine, then it passes for all builds. (https://github.com/IntelliTect/EssentialCSharp/runs/7532292828?check_suite_focus=true).

.github/workflows/Branch v9.0.yml
```diff
@@ -17,13 +17,14 @@ jobs:
...
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup dotnet SDK
-        uses: actions/setup-dotnet@v1
+        uses: actions/setup-dotnet@v2
         with:
           dotnet-version: | 
             3.1.x
             5.0.x
+            6.0.x
```
Keboo:
```suggestion
            3.1.x
            5.x
            6.x
```

global.json
```diff
@@ -0,0 +1,7 @@
+{
```
Keboo, line 1: Do we still need this after the Directory.Build.props change?

src/Chapter10/Listing10.23.RegisteringAFinalizerWithProcessExit.ps1
```diff
@@ -23,7 +23,8 @@ try {
     Get-Item "$PSScriptRoot/$ConsoleProgramProjectName" -ErrorAction Ignore | Remove-Item  -Recurse
     Set-PSDebug -Trace $traceLevel
     dotnet new Console --output "$PSScriptRoot/$ConsoleProgramProjectName"
-  
+    Set-Content ../../../../Chapter10/ProcessExitTestProgram/Directory.Build.props '<Project></Project>'
```
Keboo:
```suggestion
    Set-Content $PSScriptRoot/$ConsoleProgramProjectName/Directory.Build.props '<Project></Project>'
```

BenjaminMichaelis:
So currently we do need a global.json because we ended up with failures on dotnet 5 of "warningCS8632:Theannotationfornullablereferencetypesshouldonlybeusedincodewithina'#nullable'annotationscontext.". I have a suspicion this is because with the directory props file being empty, when the csproj file is being generated, it by default has nullability disabled. I am testing this theory by adding it to be enabled in the pwsh script here: https://github.com/IntelliTect/EssentialCSharp/pull/318 . If it passes, I will get to drop the globl.json in this pr and just instead add nullability to the pwsh script for the props file.

src/Chapter10/Listing10.23.RegisteringAFinalizerWithProcessExit.ps1
```diff
@@ -23,7 +23,8 @@ try {
     Get-Item "$PSScriptRoot/$ConsoleProgramProjectName" -ErrorAction Ignore | Remove-Item  -Recurse
     Set-PSDebug -Trace $traceLevel
     dotnet new Console --output "$PSScriptRoot/$ConsoleProgramProjectName"
-  
+    Set-Content ../../../../Chapter10/ProcessExitTestProgram/Directory.Build.props '<Project><PropertyGroup><Nullable>enable</Nullable></PropertyGroup></Project>'
```
Keboo:
```suggestion
    Set-Content ../../../../Chapter10/ProcessExitTestProgram/Directory.Build.props '<Project><PropertyGroup><Nullable>enable</Nullable></PropertyGroup></Project>'
```
```suggestion
    Set-Content $PSScriptRoot/$ConsoleProgramProjectName/Directory.Build.props '<Project><PropertyGroup><Nullable>enable</Nullable></PropertyGroup></Project>'
```

Keboo (approved):

PR #320, opened 2022-07-27
Test against different locales

### LINKED ISSUE #313: Run tests with different locales

BenjaminMichaelis:
## Summary

Fixes #313

### DIFF
--- a/.github/workflows/LocalesTest.yml
+++ b/.github/workflows/LocalesTest.yml
@@ -1,0 +1,37 @@
+name: Build and Test v9.0
+
+on:
+  push:
+    branches: [ v9.0 ]
+  pull_request:
+    branches: [ v9.0 ]
+  schedule:
+    - cron: 0 6 * * 1
+
+jobs:
+  build-and-test:
+    name: build-and-test on ${{matrix.os}}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup dotnet SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: | 
+            3.1.x
+            5.0.x
+      - name: Get Build Information
+        run: |
+          pwd
+          dotnet --info
+          dotnet --version
+      - name: Restore dependencies
+        run: dotnet restore
+      - name: Build
+        run: dotnet build --no-restore
+      - name: Test
+        run: dotnet test --no-build --no-restore --verbosity normal
--- a/.github/workflows/LocalesTest.yml
+++ b/.github/workflows/LocalesTest.yml
@@ -1,4 +1,4 @@
-name: Build and Test v9.0
+name: Test Different Locales
 
 on:
   push:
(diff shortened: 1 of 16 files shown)


.github/workflows/LocalesTest.yml
```diff
@@ -0,0 +1,65 @@
+name: Test Different Locales
+
+on:
+  push:
+    branches: [ v9.0 ]
+  pull_request:
+    branches: [ v9.0 ]
+  schedule:
+    - cron: 0 6 * * 1
```
Keboo: Do we actually need to run this on a schedule? I suspect we only care to run this when there are changes.

.github/workflows/LocalesTest.yml
```diff
@@ -0,0 +1,65 @@
...
+  
+env:
+  LOCALE_LANG: "fr_FR.UTF-8"
+
+jobs:
+  test-french-locale:
+    name: Test French Locale On Ubuntu
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
Keboo: Any reason this is set to false?

.github/workflows/LocalesTest.yml
```diff
@@ -0,0 +1,65 @@
...
+          echo "Printing date"
+          date
+      - uses: actions/checkout@v2
+      - name: Setup dotnet SDK
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: | 
+            3.1.x
+            5.0.x
+            6.0.x
```
Keboo:
```suggestion
            3.1.x
            5.x
            6.x
```

Keboo (approved):

.github/workflows/LocalesTest.yml
BenjaminMichaelis: Just for consistency with our build and test action, but no probably unecessary.

Keboo: I would posit that neither needs a cron trigger

.github/workflows/LocalesTest.yml
BenjaminMichaelis: I think as I am using a matrix currently, I am going to keep it for now while finding the bugs, but we could remove it later before merge

src/Chapter04.Tests/Listing04.07.UnexpectedInequality.Tests.cs
```diff
@@ -9,23 +9,23 @@ public class ProgramTests
...
-4.19999980926514 != 4.2
-4.2F != 4.2D";
+            string netCoreApp3expected =
+$@"{(decimal)4.2} != {(decimal)4.2000002861023}
+{(double)4.2} != {(double)4.20000028610229}
+(float){(float)4.2}M != {(float)4.2}F
+{(double)4.20000028610229} != {(double)4.2}
+{(float)4.2}F != {(double)4.2D}D
+{(double)4.19999980926514F} != {4.2D}
+{4.2F}F != {4.2D}D";
```
Keboo: We don't want to use casts here. So rather than `(float)4.2` we want to use `4.2F`. For doubles, you don't need any specific suffix, simply having a decimal point is enough.

src/Chapter04.Tests/Listing04.25.RelyingOnIndentation.Tests.cs
```diff
@@ -8,8 +8,8 @@ public class ProgramTests
         [TestMethod]
         public void MainTest()
         {
-            const string expected = 
-@"The area of the circle is: 0.00";
+            string expected = 
+$@"The area of the circle is: {0.00M}";
```
Keboo, line 12: Does this need to be a decimal rather than a double?
```suggestion
$@"The area of the circle is: {0.00}";
```

src/Chapter04.Tests/Listing04.26.IdenticalInFunctionality.Tests.cs
```diff
@@ -8,8 +8,8 @@ public class ProgramTests
         [TestMethod]
         public void MainTest()
         {
-            const string expected = 
-@"The area of the circle is: 0.00";
+            string expected = 
+$@"The area of the circle is: {0.00M}";
```
Keboo, line 12: Does this need to be a decimal rather than a double?
```suggestion
$@"The area of the circle is: {0.00}";
```

.github/workflows/LocalesTest.yml
```diff
@@ -0,0 +1,53 @@
...
+    branches: [ v9.0 ]
+  pull_request:
+    branches: [ v9.0 ]
+
+jobs:
+  set-build-and-test-locale-languages:
+    name: Build and Test ${{ matrix.locale-lang }} Locale On Ubuntu
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
BenjaminMichaelis:
```suggestion
```

src/Chapter04.Tests/Listing04.25.RelyingOnIndentation.Tests.cs
espence2003: So I tried that for both 4.25 and 4.26 and the tests failed due to actual code producing 0.00 and the change created an expected value of 0.

Issue #12, opened 2017-07-18
Tests missing


MarkMichaelis:
We are missing quite a few tests and it would be great to get the remainder in place. Rather than one person take this issue, please create a new issue titled something like "Tests missing in Chapter X' and then assign it to yourself so we know which chapters are being investigated.

If there are chapters where there are no missing tests, please still create the issue but mark it as done. :)

espence2003:
@MarkMichaelis would you still like this issue to be solved by creating issues for each chapter's tests as you stated above?

espence2003:
Listings that need tests:

- [ ] -	4.44
- [ ] -	6.12
- [ ] -	6.42
- [ ] -	6.46
- [ ] -	8.12
- [ ] -	11.07
- [ ] -	12.06
- [ ] -	12.38
- [ ] -	13.09
- [ ] -	14.01
- [ ] -	15.05
- [ ] -	16.02
- [ ] -	16.06
- [ ] -	16.09
- [ ] -	18.01
- [ ] -        18.02
- [ ] -	18.23
- [ ] -	19.02
- [ ] -	19.08
- [ ] -	19.10
- [ ] -	20.10
- [ ] -	20.13
- [ ] -	21.05
- [ ] -	21.06
- [ ] -	21.07
- [ ] -	21.08
- [ ] -	21.09
- [ ] -	21.10
- [x] -	4.15
- [x] -	4.27
- [x] -	6.1
- [x] -	6.34
- [x] -	6.49
- [x] -	10.21
- [x] -	11.06
- [x] -	11.08
- [x] -	12.15
- [x] -	12.36
- [x] -	12.37
- [x] -	13.08
- [x] -	13.12
- [x] -	13.13
- [x] -	13.16
- [x] -	13.17
- [x] -	13.18
- [x] -	13.19
- [x] -	13.21
- [x] -	13.22
- [x] -	13.23
- [x] -	14.07
- [x] -	14.08
- [x] -	14.09
- [x] -	14.10
- [x] -	14.11
- [x] -	15.01
- [x] -	15.03
- [x] -	15.04
- [x] -	15.06
- [x] -	15.07
- [x] -	15.08
- [x] -	15.09
- [x] -	15.11
- [x] -	15.15
- [x] -	15.16
- [x] -	15.17
- [x] -	15.18
- [x] -	15.21
- [x] -	15.23
- [x] -	15.29
- [x] -	16.17
- [x] -	17.02
- [x] -	17.03
[...]

PR #334, opened 2022-08-23
Add unit tests for listings in chapter 16 and 17

espence2003:
Added more unit tests needed for chapters 16 and 17. #12


src/Chapter16.Tests/Listing16.06.Tests.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    [TestClass]
+    public class ProgramTests
+    {
+        [TestMethod]
+        public void MainTest()
+        {
+            const string expected = @".\Microsoft.TestPlatform.AdapterUtilities.dll (2/15/2022 6:54:30 PM)
+.\Microsoft.TestPlatform.CommunicationUtilities.dll (5/10/2022 7:12:56 AM)
+.\Microsoft.TestPlatform.CoreUtilities.dll (5/10/2022 7:06:34 AM)
```
BenjaminMichaelis: This looks like a stack trace or something similar which is not what I think we want to be testing for as even if this works, right if there is a framework change anywhere in this stack it will fail

src/Chapter16.Tests/Listing16.09.Tests.cs
```diff
@@ -0,0 +1,48 @@
...
+.\Microsoft.TestPlatform.CommunicationUtilities.dll(103312)
+.\NuGet.Frameworks.dll(108920)
+.\testhost.exe(143776)
+.\Microsoft.VisualStudio.TestPlatform.MSTest.TestAdapter.dll(159144)
+.\Chapter16.exe(174592)
+.\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll(177056)
+.\Microsoft.VisualStudio.TestPlatform.Common.dll(181136)
+.\System.CodeDom.dll(187016)
+.\Microsoft.TestPlatform.CrossPlatEngine.dll(233872)
+.\Newtonsoft.Json.dll(468480)";
```
BenjaminMichaelis: Here as well, if an expectation is expected to be thrown might be better to check for that exception

src/Chapter16.Tests/Listing16.06.Tests.cs
espence2003: mmm okay I understand

Keboo (approved):

PR #335, opened 2022-08-23
Chapter 18 & 19 missing unit test

espence2003:
Added missing unit tests for chapters 18 and 19. #12


Keboo (approved):

src/Chapter18.Tests/Listing18.02.Tests.cs
```diff
@@ -0,0 +1,17 @@
...
+    [TestClass]
+    public class ProgramTests
+    {
+        [TestMethod]
+        public void MainTest()
+        {
+            const string expected = @"";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(
+                expected, Program.Main);
```
BenjaminMichaelis: What is the goal of this test? Is it supposed to return nothing or is there something better we can test?

src/Chapter19.Tests/Listing19.06.Tests.cs
```diff
@@ -0,0 +1,17 @@
...
+    [TestClass]
+    public class ProgramTests
+    {
+        [TestMethod]
+        public void MainTest()
+        {
+            const string expected = @"ERROR: Operation is not valid due to the current state of the object.";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(
+                expected, Program.Main);
```
BenjaminMichaelis, line 14: Is this an exception that we want to catch or an error we are printing to the console?

src/Chapter18.Tests/Listing18.02.Tests.cs
espence2003: I looked at the listing 18.02 and I don't think we actually need a unit test for it

PR #336, opened 2022-08-23
Added Missing unit tests for chapters 20 and 21

espence2003:
Created the missing unit tests for these chapters. #12

### DIFF
--- a/src/Chapter20.Tests/Listing20.07.Tests.cs
+++ b/src/Chapter20.Tests/Listing20.07.Tests.cs
@@ -1,16 +1,15 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
-using System.Threading.Tasks;
+﻿//using Microsoft.VisualStudio.TestTools.UnitTesting;
+//using System.Threading.Tasks;
 
-namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter20.Listing20_07.Tests
-{
-    [TestClass]
-    public class ProgramTests
-    {
-        [Ignore]
-        [TestMethod]
-        public async Task MainTest()
-        {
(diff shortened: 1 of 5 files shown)


src/Chapter21.Tests/Listing21.04.Tests.cs
```diff
@@ -0,0 +1,26 @@
...
+            const string expected = @"ERROR: AggregateException:
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.
+  UnauthorizedAccessException - Attempted to perform an unauthorized operation.";
```
BenjaminMichaelis: We probably want to catch and check this exception instead of checking the console output

src/Chapter20.Tests/Listing20.07.Tests.cs
```diff
@@ -0,0 +1,17 @@
...
+    [TestClass]
+    public class ProgramTests
+    {
+        [TestMethod]
+        public void MainTest()
+        {
+            const string expected = @"";
+
+            IntelliTect.TestTools.Console.ConsoleAssert.Expect(
+                expected, AsyncEncryptionCollection.Main);
```
BenjaminMichaelis: What are we testing here? Is the goal of the program to return an empty string?

Keboo (approved):

PR #364, opened 2022-09-26
Refactored out into the Shared/Tests solution directory

MarkMichaelis:
Refactored out a file into the Shared/Tests solution directory that may be used to execute PowerShell scripts.  This is used to execute the HelloWorldInC#.ps1 file.

Fixes #209

PR #366, opened 2022-10-01
Completed Chapter 1, handling Top-Level Statements a for initial set of listings.

MarkMichaelis:


Keboo (commented):
This looks like it contains a bunch of the changes from #364 can we separate these?

MarkMichaelis:
I have closed PR #364. This PR supersedes it.

src/Chapter10.Tests/Listing10.23.RegisteringAFinalizerWithProcessExit.Tests.cs
```diff
@@ -1,41 +1,39 @@
...
-
     public class DisposeTests
     {
+        public TestContext TestContext { get; set; } = null!; // Auto-initialized by MSTest.
 
-        static string Ps1Path { get; } = Path.GetFullPath("../../../../Chapter10/Listing10.23.RegisteringAFinalizerWithProcessExit.ps1", Environment.CurrentDirectory);
-        static bool PowerShellNotAvailable = PowerShellTestsUtilities.PowerShellNotInstalled();
-        static string PowershellEnvironmentVariableName { get; set; } = "powershell";
+        static string Ps1DirectoryPath { get; } =
+            Path.GetFullPath(Path.Join("..", "..", "..", "..", "Chapter10"), Environment.CurrentDirectory);
```
Keboo: I typically hate relative paths like this as they make assumptions on the working directory. We have a solution in the other repo that instead of counting a number of parent directories, it instead looks up through the directories for the ".git" folder to find the repo root and builds relative paths based on that.

BenjaminMichaelis: https://www.nuget.org/packages/IntelliTect.Multitool/#readme-body-tab

src/Shared/Tests/PowerShellTestUtilities.cs~Minor rename of PowerShellTestUtilities
```diff
@@ -0,0 +1,77 @@
+using System;
```
Keboo: Did you mean to include this file?

src/Chapter10.Tests/Listing10.23.RegisteringAFinalizerWithProcessExit.Tests.cs
MarkMichaelis: You are right.  I eliminated the problem in the PowerShell script but not the C# code. @BenjaminMichaelis has been suggesting I use the common library for this but I was reluctant to take an additional library dependency, even in a test. However, your point about working directory is significant.

src/Shared/Tests/PowerShellTestUtilities.cs~Minor rename of PowerShellTestUtilities
MarkMichaelis: No, but interesting... What created it?

src/Chapter10.Tests/Listing10.23.RegisteringAFinalizerWithProcessExit.Tests.cs
MarkMichaelis: Leveraged Multitool as @BenjaminMichaelis  suggested.

Directory.Build.props
```diff
@@ -1,15 +1,16 @@
...
     <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
     <ProductName Condition="'$(ProductName)'==''">Chapter$(ChapterNumber)</ProductName>
     <TreatWarningsAsErrors Condition="'$(TreatWarningsAsErrors)'==''">false</TreatWarningsAsErrors>
     <Authors>Mark Michaelis</Authors>
     <Company>Addison Wesley</Company>
     <Copyright>Copyright © Addison Wesley 2022</Copyright>
     <Product>Essential C# $(LangVersion)</Product>
     <Title>$(Product)</Title>
-    <Version>9.0</Version>
```
BenjaminMichaelis, line 12: Do we need both a "version" and a "langVersion" or can these be combined or one referenced from the other

src/Chapter01.Tests/Listing01.01.Tests.cs
```diff
@@ -1,18 +1,60 @@
...
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter01.Listing01_01.Tests
 {
     [TestClass]
     public class HelloWorldTests
     {
+        static string Ps1DirectoryPath { get; } =
+            Path.GetFullPath(Path.Join("..", "..", "..", "..", "Chapter01"), Environment.CurrentDirectory);
+        static string Ps1Path { get; } = 
+            Path.GetFullPath(
+                Path.Join(Ps1DirectoryPath, "Listing01.01.HelloWorldInC#.ps1"), Environment.CurrentDirectory);
```
BenjaminMichaelis: We could use the MultiTool to normalize the paths here as well

src/Chapter01.Tests/Listing01.02B.Tests.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter01.Listing01_06.Tests
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter01.Listing01_02B.Tests
```
BenjaminMichaelis, line 3: Do we want to merge this in before running the listing manager or wait for the word docs to get updated?

src/Chapter01/Listing01.01.HelloWorldInC#.ps1
```diff
@@ -22,21 +23,24 @@ try {
...
 "@
 
     $codeListing > Program.cs
     Get-Content Program.cs
-    dotnet run
+
+    Write-Warning "Remove dotnet build step once we upgrade to version 10"
+    # Build with version 10 because dotnet version 6 defaults to an ImplicitUsings element in the project file.
+    dotnet build /property:LangVersion=10
+    dotnet run --no-build
```
BenjaminMichaelis:
```suggestion
    dotnet run
```

Aren't we in v11 in this pr now? Should this be updated to remove the build step and remove the warning?

src/Chapter01/Listing01.04.BreakingApartHelloWorld.cs
```diff
@@ -1,12 +1,13 @@
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter01.Listing01_04
 {
     #region INCLUDE
-    public class Program//Class definition
+    public class Program                // BEGIN Class definition
```
BenjaminMichaelis, line 4: We should check how this formats in the manuscript, I wonder what this might look like

src/Chapter02/Listing02.13.UsingTheNewlineCharacterToInsertANewline.cs
```diff
@@ -1,14 +1,14 @@
...
 {
-    #region INCLUDE
     public class DuelOfWits
     {
         public static void Main()
         {
+            #region INCLUDE
             System.Console.Write(
                 "\"Truly, you have a dizzying intellect.\"");
             System.Console.Write("\n\"Wait 'til I get going!\"\n");
```
BenjaminMichaelis, line 10: Do we want to use `Environment.NewLine` here or is it still too early in the book? Or just use Console.WriteLine maybe?
```suggestion
            System.Console.WriteLine("\"Wait 'til I get going!\"");
```

src/Chapter02/Listing02.14B.RawLiteralStrings.cs
```diff
@@ -0,0 +1,28 @@
...
+//      / i /   /\   \
+//     / b /   /  \   \
+//    / o /   /\   \   \
+//   / m /   /  \   \   \
+//  /___/___/____\   \   \
+// /   /          \   \  /\
+///___/____________\___\/  \
+//\   \             \   \  /
+// \___\_____________\___\/
+//end""");
```
BenjaminMichaelis:
```suggestion
            System.Console.Write(
                """begin
           ____
          /   /\
         /___/  \
        / s /\  /\
       / u /  \/  \
      / i /   /\   \
     / b /   /  \   \
    / o /   /\   \   \
   / m /   /  \   \   \
  /___/___/____\   \   \
 /   /          \   \  /\
/___/____________\___\/  \
\   \             \   \  /
 \___\_____________\___\/
end""");
```
Is this intentionally commented out?

src/Chapter10/Listing10.23.RegisteringAFinalizerWithProcessExit.ps1
```diff
@@ -1,67 +1,79 @@
...
-    $codeListing > "$PSScriptRoot/$ConsoleProgramProjectName/Program.cs"
-    Get-Content "$PSScriptRoot/$ConsoleProgramProjectName/Program.cs"  # Display the listing
+switch ($TestStage) {
+    "create" {
+        try {
+            Write-Host "`$projectDirectory: $projectDirectory"
+            Get-Item $projectDirectory -ErrorAction Ignore | Remove-Item  -Recurse -Force
+            Set-PSDebug -Trace $TraceLevel
+            dotnet new console --output $projectDirectory
+            Set-Content (Join-Path $projectDirectory 'Directory.Build.props') '<Project><PropertyGroup><Nullable>enable</Nullable></PropertyGroup></Project>'
```
BenjaminMichaelis: Is nullable enabled by default and we don't need to explicitly put that in the proj? However it doesn't hurt so we can just leave it.

src/Chapter11/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "WSL": {
+      "commandName": "WSL2",
+      "distributionName": ""
+    }
+  }
+}
```
BenjaminMichaelis: Do we want to be committing this file?

src/Shared/Tests/PowerShellTestUtilities.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace AddisonWesley.Michaelis.EssentialCSharp.Shared.Tests
+{
+    public class PowerShellTestUtilities
+    {
+        public static bool WindowsEnvironment()
+        {
+            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        }
+
```
BenjaminMichaelis, line 14:
```suggestion
```
nit: extra line

src/Shared/Tests/PowerShellTestUtilities.cs
```diff
@@ -0,0 +1,69 @@
...
+            powerShell.StartInfo.Arguments = $"-noprofile -command \"{scriptPath} {arguments}\"";
+            powerShell.Start();
+            psOutput = powerShell.StandardOutput.ReadToEnd();
+            powerShell.WaitForExit();
+
+            return powerShell.ExitCode;
+        }
+
+        public static string? PowerShellCommand => _PowerShellCommand.Value;
+        public static bool PowerShellNotInstalled =>_PowerShellCommand.Value == null;
```
BenjaminMichaelis:
```suggestion
        public static bool PowerShellNotInstalled => _PowerShellCommand.Value is null;
```
nit: add a space, and can this be is null?

BenjaminMichaelis (commented):
I would double check all the warnings your getting are expected

src/Chapter01.Tests/Listing01.02B.Tests.cs
MarkMichaelis: I don't know what the order of operations is.

src/Chapter01/Listing01.04.BreakingApartHelloWorld.cs
MarkMichaelis: I assume it looks like a comment. I'm not sure I understand what you are asking.

src/Chapter02/Listing02.14B.RawLiteralStrings.cs
MarkMichaelis: Nope... thanks for noticing.

src/Chapter01/Listing01.01.HelloWorldInC#.ps1
MarkMichaelis: Agreed!

Directory.Build.props
MarkMichaelis: Interesting... The history is ambiguous. I'm not seeing where `Version` is used anymore.  When `LangVersion` was added `Version` was kept but I'm not sure why. I have removed `Version`.
Good catch!!

src/Chapter02/Listing02.13.UsingTheNewlineCharacterToInsertANewline.cs
MarkMichaelis: The text is explicit showing the difference between `WriteLine` and `Write`.

global.json
```diff
@@ -1,7 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.0",
-    "rollForward": "latestMinor",
-    "allowPrerelease": false
+    "rollForward": "latestPatch",
```
Keboo: Any reason why the version was dropped? This seems problematic as people will not get errors if they don't have 7.0 installed.

.gitignore
```diff
@@ -1,5 +1,7 @@
 # Files specific to this project:
 *.encrypt  # Generated by one of the projects and left around on occasion.
+src/Chapter10/ProcessExitTestProgram/
+HelloWorld/
```
Keboo: Is this expected to cover the HelloWorld program created by the powershell script in chapter 1? If so there is likely a working directory issue. This is filtering from the root directory, while the chapter 1 csproj looks for a HelloWorld directory inside of it. 
I would suggest that we do something to the path's similar to C# files. Perhaps <dir>.g? to indicate that a folder is generated, then we can have a singular globbing pattern. 

`**/*.g/` for this file
`<Compile Remove="**\*.g\**" />` to be put in the Directory.Build.targets file

src/Chapter01/Chapter01.csproj
```diff
@@ -8,9 +8,13 @@
     <!-- CS8600: Trust that null assignment doesn't cause 
     unexpected exception as concept to check for null hasn't been explained yet-->
     <IgnoredWarnings>$(IgnoredWarnings),CS8600,CA1052,IDE0059,IDE0060</IgnoredWarnings>
+    <StartupObject>AddisonWesley.Michaelis.EssentialCSharp.Shared.Program</StartupObject>
   </PropertyGroup>
   <Import Project="..\Chapter.props" />
   <ItemGroup>
+    <Compile Remove="HelloWorld\**" />
+    <Compile Remove="Listing01.01.HelloWorldInC#.cs" />
+    <Compile Remove="Listing01.01A.HelloWorldInC#.cs" />
```
Keboo, line 17: This one is removed and not added to the `None` set like `Listing01.01.HelloWorldInC#.cs` is this intentional?

src/Chapter16.Tests/Listing16.15.Tests.cs
```diff
@@ -1,22 +1,98 @@
...
 {
     [TestClass]
     public class ProgramTests
     {
+        public TestContext TestContext { get; set; } = null!; // Assigned by MSTest.
+
         [TestMethod]
         public void SelectingAnonymousTypeFollowingGroupClause()
         {
+            string e = Shared.NetCore.GetNetCoreVersion() switch
```
Keboo: This entire switch statement appears unused and duplicated the one below.

src/Chapter16.Tests/Listing16.15.Tests.cs
```diff
@@ -1,22 +1,98 @@
...
+                    FirstOrDefault, GroupBy, GroupJoin, Intersect, IntersectBy, Join, 
+                    Last, LastOrDefault, LongCount, Max, MaxBy, Min, MinBy, OfType, 
+                    Order, OrderBy, OrderByDescending, OrderDescending, Prepend, Range, 
+                    Repeat, Reverse, Select, SelectMany, SequenceEqual, Single, 
+                    SingleOrDefault, Skip, SkipLast, SkipWhile, Sum, Take, TakeLast, 
+                    TakeWhile, ThenBy, ThenByDescending, ToArray, ToDictionary, ToHashSet, 
+                    ToList, ToLookup, TryGetNonEnumeratedCount, Union, UnionBy, Where, Zip, 
+                    """;
+
+            TestContext.WriteLine(Shared.NetCore.GetNetCoreVersion());
```
Keboo: This code appears to invoke `Shared.NetCore.GetNetCoreVersion()` several times. Can we cache the return in a local.
```suggestion
            string dotnetVersion = Shared.NetCore.GetNetCoreVersion(); 
            TestContext.WriteLine(dotnetVersion);
```

src/Chapter16.Tests/Listing16.15.Tests.cs
```diff
@@ -1,22 +1,98 @@
...
+            if (string.Compare(Shared.NetCore.GetNetCoreVersion(), "7") < 0)
```
Keboo: Is there a reason these are using `string.Compare` rather than `string.StartsWith`?

.github/workflows/build-and-test.yml
```diff
@@ -19,12 +19,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Setup dotnet SDK
-        uses: actions/setup-dotnet@v2
+        uses: actions/setup-dotnet@v3
```
Keboo, line 22: dependabot update has just merged that will keep these up to date.

BenjaminMichaelis: Might of been me when I was helping fix the action on his computer for 7.0 that I bumped it at the same time...

Directory.Build.props
Keboo: Version is used by `dotnet build`, `pack`, and `publish` to set the output version numbers. It, along with a few other properties and the means to control the output version. Given that none of these projects are expected to be done that way  I think it makes sense to omit it.

MarkMichaelis: Thanks... I was pretty sure it wasn't added by mistake but couldn't reproduce the need when I tested it.  Glad to know removing it wasn't a mistake.

src/Chapter16.Tests/Listing16.15.Tests.cs
MarkMichaelis: Indeed... incomplete refactoring. :)

MarkMichaelis: Indeed... incomplete refactoring. :)

src/Chapter01/Chapter01.csproj
MarkMichaelis: @Keboo: Yes... you cannot specify /main (used to run precode via which a user selects which listing to run) with top-level statements. :(.  Alternative suggestions welcome.

MarkMichaelis: Once web edit/run is supported this becomes irrelevant so ignoring.

src/Chapter01/Listing01.08.NoIndentationFormatting.cs
```diff
@@ -5,7 +5,7 @@ public class HelloWorld
 {
 public static void Main()
 {
-System.Console.WriteLine("Hello Inigo Montoya");
+            Console.WriteLine("Hello Inigo Montoya");
```
Keboo: Given this is the no indentation example, do we really want this indentation?

src/Chapter01/Listing01.09.RemovingWhitespace.cs
```diff
@@ -2,6 +2,6 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter01.Listing01_09
 {
     #region INCLUDE
     public class HelloWorld{public static void Main()
-{System.Console.WriteLine("Hello Inigo Montoya");}}
+{ Console.WriteLine("Hello Inigo Montoya");}}
```
Keboo: Inconsistent spacing and too many closing braces
```suggestion
{ Console.WriteLine("Hello Inigo Montoya"); }
```

src/Chapter10.Tests/Listing10.23.RegisteringAFinalizerWithProcessExit.Tests.cs
```diff
@@ -1,41 +1,43 @@
...
 
         [ClassCleanup]
         public static void RemoveProcessExitProj()
         {
-            if (PowerShellNotAvailable) { return; }
-            string testStatus = "cleanup";
-            Process powershell = Process.Start(PowershellEnvironmentVariableName, $"-noprofile -command \"{Ps1Path} 0 null {testStatus}\"");
-            powershell.WaitForExit();
+            Assert.AreEqual<int>(0, RunPowerShellScript(
+                "cleanup", out string _));
```
Keboo: This assertion doesn't actually work.
By the time the cleanup method gets invoked the all the tests in the class have already reported success/failure.

image

.gitignore
```diff
@@ -1,5 +1,7 @@
 # Files specific to this project:
 *.encrypt  # Generated by one of the projects and left around on occasion.
+**/*.gc/ # Folders generated buy tests will use the *.g extension so they aren't check in.
```
Keboo:
```suggestion
**/*.g/ # Folders generated buy tests will use the *.g extension so they aren't check in.
```

src/Chapter10.Tests/Listing10.23.RegisteringAFinalizerWithProcessExit.Tests.cs
MarkMichaelis: Agreed... Also, if debugging and you stop, cleanup won't run.  Removing the assert and adding cleanup to classinitialize as well.

Keboo (approved):

src/Chapter01.Tests/Listing01.01.Tests.cs
```diff
@@ -1,18 +1,61 @@
...
+                        "The template \"Console App\" was created successfully.",
+                        "Build succeeded.",
+                        "0 Error(s)",
+                        "0 Warning(s)"
+                    };
+
+                foreach (string text in expectedText)
+                {
+                    Assert.IsTrue(psOutput.Contains(text),
+                        $"Unexpectedly, '{text}' did not appear in the console output: \n {psOutput}");
```
BenjaminMichaelis:
```suggestion
                        $"Unexpectedly, '{text}' did not appear in the console output: {Environment.NewLine} {psOutput}");
```

BenjaminMichaelis (approved):

PR #385, opened 2022-10-31
Renumbered code files in Chapter 2

MarkMichaelis:


src/Chapter02.Tests/Listing02.18.Tests.cs
```diff
@@ -1,19 +1,19 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter02.Listing02_18.Tests
 {
     [TestClass]
     public class UppercaseTests
     {
         [TestMethod]
-        public void Main_InputLorem_OutputIsCapitalized()
+        public void Main_InputLorem_OutputNotCapitalized()
```
Keboo, line 9: Something seems off here. The text string below seems to show the same string going in and out, but this test name implies "NotCapitalized" which I would have taken to mean lowercase, but that doesn't match the assertion. Perhaps simply rename to "Unchanged"?

Keboo (approved):

src/Chapter02.Tests/Listing02.18.Tests.cs
MarkMichaelis: The syntax, text.ToUpper() would indicate that the text is changed to uppercase.  However, the test shows that no such casing change occurs.  I'm open to suggestions.

Keboo (approved):

PR #388, opened 2022-11-02
Switched from cs to XML on csproj files

MarkMichaelis:
These files are not used as project files, just sample files that are listed in the chapter.


Keboo (approved):

src/Chapter03/Listing03.03.EnablingNullableProjectWide.xml
```diff
@@ -1,9 +1,7 @@
-#region INCLUDE
```
BenjaminMichaelis, line 1: We can't not have regions, but it can be an xml file

BenjaminMichaelis (changes requested):

BenjaminMichaelis (approved):

PR #389, opened 2022-11-03
Remove obsolete references to System.ValueTuple

MarkMichaelis:


Keboo (approved):
Change looks fine, but it also looks like it contains your csproj changes that have already merged. You probably want to rebase first.

PR #391, opened 2022-11-05
Updated source code for Chapter 3.

MarkMichaelis:


src/Shared/CompilerAssert.cs
```diff
@@ -68,52 +64,60 @@ public bool Equals(CompileError rhs)
...
-            }
+        }
 
-            async static public Task<CompileError[]> ExpectErrorsAsync(
+        async static public Task<CompileError[]> ExpectErrorsAsync(
             string sourceCode, params CompileError[] diagnostics)
         {
+            CompileError[] actualCompileErrors = await CompileAsync(sourceCode);
+
+            if (diagnostics is { } && diagnostics.Length > 0)
```
Keboo:
```suggestion
            if (diagnostics?.Length > 0)
```

src/Shared/CompilerAssert.cs
```diff
@@ -68,52 +64,60 @@ public bool Equals(CompileError rhs)
...
+                        $"The expected Ids do not match for item {i}: " +
+                        $"{diagnostics[i].Id}: {diagnostics[i].Message} <> {actualCompileErrors[i].Id}: {actualCompileErrors[i].Message}");
+                }
+                return actualCompileErrors;
+            }
+
+            Assert.Fail("The expected compilation errors did not occur.");
+            return null!;  // This code will never execute due to the Fail line above.
+        }
+        async static public Task<CompileError[]> CompileAsync(
```
Keboo:
```suggestion
        public static async Task<CompileError[]> CompileAsync(
```

src/Shared/CompilerAssert.cs
```diff
@@ -41,8 +37,8 @@ public override string ToString()
         // override object.Equals
         public override bool Equals(object? obj)
         {
-            return (obj is CompileError)
-              && Equals((CompileError)obj);
+            return (obj is CompileError error)
+              && Equals(error);
```
Keboo, line 41: Making this a record strut would eliminate most of the code for `CompileError`

src/Shared/Tests/Parallel.runsettings
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Keboo, line 1: Are we configuring something to use this file?

Keboo (approved):

src/Shared/CompilerAssert.cs
MarkMichaelis: Agreed, perhaps even a record struct. I chose not to take on the distraction in spite of the temptation. :)

src/Shared/Tests/Parallel.runsettings
MarkMichaelis: I'd like to have it used by default when running tests but my understanding is that it requires a reference in the command line (or perhaps an environment variable works?).  Suggestions welcome.

src/Shared/CompilerAssert.cs
MarkMichaelis: Agreed.... there wasn't enough value in the change so I avoided the distraction. See #393.

src/Shared/Tests/Parallel.runsettings
Keboo: I think MSBuild properties are you best bet. https://learn.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2022#set-a-build-property

PR #396, opened 2022-11-07
Chapter 4 updates

MarkMichaelis:
This does not include review of any of the nullability code samples (Listing 37-39)

PR #397, opened 2022-11-12
Code Completed for Chapter 4 Technical Review

MarkMichaelis:


BenjaminMichaelis:
Does this supersede #396?

src/Chapter04/Listing04.22.NestedIf.cs
```diff
@@ -0,0 +1,48 @@
...
+            Console.Write(
+                "What is the maximum number " +
+                "of turns in tic-tac-toe?" +
+                " (Enter 0 to exit.): ");
+
+            // int.Parse() converts the ReadLine()
+            // return to an int data type
+            input = int.Parse(Console.ReadLine());
+
+            // Check 1.
```
BenjaminMichaelis, line 21:
```suggestion
            // Condition 1.
```

nit: keep verbiage the same throughout listing

src/Chapter04/Listing04.22.NestedIf.cs
```diff
@@ -0,0 +1,48 @@
+// Justification: Checkihng for null isn't discussed yet.
```
BenjaminMichaelis:
```suggestion
// Justification: Checking for null isn't discussed yet.
```

src/Chapter04/Listing04.22.NestedIf.cs
BenjaminMichaelis:
```suggestion
            // Condition 1.
```

nit: keep verbiage the same throughout listing

src/Chapter04/Listing04.38.NullConditionalOperator.cs
```diff
@@ -1,35 +1,39 @@
...
 
             int? length = segments?.Length;
-            if (length is { } && length != 0)
+            #region EXCLUDE
+            // Pattern matching also allows:
+            // the following but this is not used
+            // until covering the topic fully in Chapter 7.
+            if (length is not null and not 0) { /*...*/ }
+            #endregion EXCLUDE
+            if (length is not null && length != 0)
```
BenjaminMichaelis, line 24: isn't length not null here? Removing the necessity of the following two not null checks and int being nullable?

src/Chapter04/Listing04.20.ConstantInterpolatedStrings.cs
```diff
@@ -0,0 +1,21 @@
+#pragma warning disable CS0168 // Variable is declared but never used
+#pragma warning disable CS0219 // Variable is assigned but its value is never used
+#pragma warning disable IDE0059 // Unnecessary assignment of a value
```
BenjaminMichaelis: I feel like I keep seeing a lot of unnecessary suppressions, like here I don't think I see anything we don't use. And I think there's other cases of this as well that I'd suggest we look for.

src/Chapter04/Listing04.49.InputValidation.cs
```diff
@@ -31,7 +31,7 @@ public static void Main(params string[] args)
             }
             else
             {
-                System.Console.WriteLine(
+                Console.WriteLine(
                   "\nERROR:  Enter a value from 1-9. "
                   + "Push ENTER to quit");
```
BenjaminMichaelis:
```suggestion
                  $"{ Environment.NewLine }ERROR:  Enter a value from 1-9. "
                  + "Push ENTER to quit");
```
can we use environment.newline over \n here?

src/Chapter04/Listing04.50.SwitchStatement.cs
```diff
@@ -36,7 +36,7 @@ public static void Main()
                 default:
                     // If none of the other case statements
                     // is encountered then the text is invalid
-                    System.Console.WriteLine(
+                    Console.WriteLine(
                     "\nERROR:  Enter a value from 1-9. "
```
BenjaminMichaelis:
```suggestion
                    $"{ Environment.NewLine }ERROR:  Enter a value from 1-9. "
```
And here

src/Chapter04/Table04.04.CheckForNull.cs
```diff
@@ -2,44 +2,98 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter04.Table04_04
...
+
+        }
+        public static void IsObject(string? uri = null!)
+        {
+            // 4.
+            int number = 0;
+            if ((uri is object)
+                // Warning CS0183: The given
+                // expression is always not
+                // null.
```
BenjaminMichaelis, line 56: Github actions seems to suggest that the error message is "The given expression is always of the provided ('object') type", is maybe this error message slightly off?

BenjaminMichaelis (changes requested):

src/Chapter04/Listing04.38.NullConditionalOperator.cs
MarkMichaelis: No, `length` would be null here if `segments` is null.

BenjaminMichaelis (approved):

Keboo (approved):

PR #409, opened 2022-12-02
Chapter 5 tests

ZackBWard:
fixes https://github.com/IntelliTect/EssentialCSharpManuscript/issues/589

### DIFF
--- a/src/Chapter05.Tests/Listing05.31.Tests.cs
+++ b/src/Chapter05.Tests/Listing05.31.Tests.cs
@@ -5,4 +5,22 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_31.Tests;
 [TestClass]
 public class ProgramTests
 {
+    [TestMethod]
+    public void Main_NoArgs_ExpectErrors()
+    {
+        string expected = "Value cannot be null. (Parameter \'httpsUrl\')";
+
+        IntelliTect.TestTools.Console.ConsoleAssert.Expect(expected,
+        () =>
+        {
+            try
+            {
+                Program.Main();
+            }catch(Exception ex)
+            {
+                Console.WriteLine(ex.Message);
+            }
+        });
+    }
+
 }
--- a/src/Chapter05.Tests/Listing05.32.Tests.cs
+++ b/src/Chapter05.Tests/Listing05.32.Tests.cs
@@ -5,5 +5,22 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_32.Tests;
 [TestClass]
 public class ProgramTests
 {
+    [TestMethod]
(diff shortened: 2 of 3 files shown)


src/Chapter05.Tests/Listing05.31.Tests.cs
```diff
@@ -5,4 +5,22 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_31.Tests;
...
+        IntelliTect.TestTools.Console.ConsoleAssert.Expect(expected,
+        () =>
+        {
+            try
+            {
+                Program.Main();
+            }catch(Exception ex)
+            {
+                Console.WriteLine(ex.Message);
+            }
```
Keboo, line 22: This will suppress all exceptions. The test name seems to indicate that we expect an exception. If that is the case we will want the test to fail if there are not exception. Right now this test will likely work in all cases (even the failure case).

src/Chapter05.Tests/Listing05.32.Tests.cs
```diff
@@ -5,5 +5,22 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_32.Tests;
...
+        string expected = "Value cannot be null. (Parameter \'httpsUrl\')";
 
+        IntelliTect.TestTools.Console.ConsoleAssert.Expect(expected,
+        () =>
+        {
+            try
+            {
+                Program.Main();
+            }
+            catch (Exception ex)
```
Keboo, line 20: Same comment here, we don't want to suppress the failure cases.

src/Chapter05.Tests/Listing05.33.Tests.cs
```diff
@@ -5,5 +5,180 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter05.Listing05_33.Tests;
...
+        string expected = "Value cannot be null. (Parameter \'httpsUrl\')";
+
+        IntelliTect.TestTools.Console.ConsoleAssert.Expect(expected,
+        () =>
+        {
+            try
+            {
+                Program.Main(args);
+            }
+            catch (Exception ex)
```
Keboo, line 38: Same comment here, we don't want to suppress the failure cases.

src/Chapter05.Tests/Listing05.31.Tests.cs
ZackBWard: Won't it not write to console causing it to fail?

Keboo: O I get what you are doing now. Yes this will work, is just doing string comparisons so different Exception types might not be the same, but I am fine with this. Disregard my prior comments, this does work. I have an issue logged on the ConsoleAssert class to make a real method to assert on exceptions.

Keboo (approved):

PR #414, opened 2022-12-09
add wrapper file so 1.1 can be run

twofingerrightclick:
## Summary

uses #line directive to fake the appearance of running 1.1
Includes explanation if a reader were to investigate


src/Chapter01/TopLevelStatementWrapperForListing01.01.cs
```diff
@@ -0,0 +1,16 @@
+// See https://aka.ms/new-console-template for more information
```
Keboo, line 1:
```suggestion
```

Keboo (approved):

src/Chapter01/TopLevelStatementWrapperForListing01.01.cs
MarkMichaelis: This was generated by the project so I suspect we should leave it.

PR #443, opened 2023-01-26
Proposed cleanup of list 7.32

Keboo:
## Summary

Cleanup of listing 7.32


BenjaminMichaelis (approved):
Still might want @MarkMichaelis opinion, but looks fine to me

PR #444, opened 2023-01-26
Add new Chapter 2 listings and update tests

peter-intellitect:
Added new listings.
Updated listing tests in visual studio.


src/Chapter02/example.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter02.Listing02_01
```
Keboo: Should this file be renamed to match its listing number?

src/Chapter02.Tests/Listing02.17.Tests.cs
```diff
@@ -1,18 +1,26 @@
...
 
-        IntelliTect.TestTools.Console.ConsoleAssert.Expect(
-            expected, PalindromeLength.Main);
+        string expected = @$"Enter your first name: 
+<<{firstName}
+>>Enter your last name: 
+<<{lastName}
+>>Your full name is:
+{firstName} {lastName}";
+
```
BenjaminMichaelis, line 21: Nit: extra unnecessary space
```suggestion
{firstName} {lastName}";
```

src/Chapter02.Tests/Listing02.21.Tests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        try
+        {
+            JsonDocument.Parse(result);
+
+        }
+        catch (JsonException ex)
+        {
+            string message = "Invalid JSON format: " + ex.Message;
+            Assert.Fail(message);
```
BenjaminMichaelis: Can we just make this one line instead of assigning a variable?
```suggestion
            Assert.Fail("Invalid JSON format: " + ex.Message);
```

src/Chapter02.Tests/Listing02.21.Tests.cs
```diff
@@ -0,0 +1,25 @@
...
+    public void Main_OutputsJson()
+    {
+        string result = IntelliTect.TestTools.Console.ConsoleAssert.Execute("", Program.Main);
+
+        try
+        {
+            JsonDocument.Parse(result);
+
+        }
+        catch (JsonException ex)
```
BenjaminMichaelis: Do we just want to catch a JsonException or should we catch any exception?

src/Chapter02/Listing02.17.CombiningStringLiteralsWithInterpolation.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class Program
+    {
+        public static void Main()
+        {
+            string firstName;
+            string lastName;
+
+           
+            Console.WriteLine("Enter your first name: ");
```
BenjaminMichaelis:
```suggestion
            string lastName;
           
            Console.WriteLine("Enter your first name: ");
```

Nit: spacing

src/Chapter02/Listing02.21.WorkingWithJSONUsingRawStringLiterals.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public static void Main()
+        {
+            string firstName = "Forest";
+
+            string lastName = """Gump"""; // Single-line raw string literal.
+
+            string greeting = // Single-line raw string literal with interpolation.
+
+            $"""Hello, I'm {firstName}. {firstName} {lastName}""";
```
BenjaminMichaelis: Do we want these comments here or should they be one line above? And then the string itself can be moved up? A few of these in this document

src/Chapter02.Tests/Listing02.21.Tests.cs
Keboo: I would propose that we don't catch any exception and let the test fail since that is what it will do anyway

src/Chapter02/Listing02.21.WorkingWithJSONUsingRawStringLiterals.cs
MarkMichaelis: We can't have lines longer than 72 characters, so this should be on the prior line.

src/Chapter02.Tests/Listing02.17.Tests.cs
```diff
@@ -1,18 +1,26 @@
...
+
+        IntelliTect.TestTools.Console.ConsoleAssert.Expect(expected,
```
BenjaminMichaelis:
```suggestion
{firstName} {lastName}";

        IntelliTect.TestTools.Console.ConsoleAssert.Expect(expected,
```
Nit: extra unnecessary space

twofingerrightclick: I hate manually formatting code haha. I always use an opinionated linter (but that would change the whole file)... I will get these extra lines.

BenjaminMichaelis (approved):

PR #448, opened 2023-02-01
Chapter 10 source code

MarkMichaelis:


src/Chapter09.Tests/Listing09.06.CustomizingARecord.Tests.cs
```diff
@@ -0,0 +1,35 @@
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter09.Listing09_06.Tests;
+
+[TestClass]
+public class AngleTests
+{
+    [TestMethod]
+    public void Title_Null_ThrowsArgumentNullException() => _ = new Angle(0,0,0);
```
Keboo: I am not sure what this test is intending to do?
The name implies you expect and ArgumentNullException, but this test is not testing for exceptions. There is also not a Title property, just a Name property, but it doesn't do any null checks. What was the goal here?

Keboo (commented):
I pushed up a commit with most of my comments resolved. However there are a few things I am not sure how to handle

src/Chapter09.Tests/Listing09.04.DeclaringRecordClass.Tests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+[TestClass]
+public partial class CoordinateTests
+{
+    [TestMethod]
+    public void Create_Coordinate_IsReadOnly()
+    {
+        Coordinate coordinate = new(
+            new Angle(180, 0, 0), new Angle(180, 0, 0));
+    }
```
Keboo, line 14: This test content does not actually test that a coordinate is readonly. That seems like something we would test with a CompilerAssert. Is that what is intended here?

src/Chapter09.Tests/Listing09.06.CustomizingARecord.Tests.cs
MarkMichaelis: Deleted

src/Chapter09.Tests/Listing09.04.DeclaringRecordClass.Tests.cs
MarkMichaelis: This needs to be a CompilerAssert test.

MarkMichaelis: I have deleted it for the time being since CompilerAssert doesn't currently support testing a code snippet in the test.

MarkMichaelis:
Merged with Chapter 9

PR #461, opened 2023-02-13
Added synchronization when running PowerShell scripts

MarkMichaelis:
Previously, these would frequently fail on the local machine due to simultaneous execution across target frameworks.


src/Chapter10.Tests/Listing10.17.RegisteringAFinalizerWithProcessExit.Tests.cs
```diff
@@ -6,8 +6,10 @@
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter10.Listing10_17.Tests;
 
 [TestClass]
-public class DisposeTests
+public partial class DisposeTests
 {
+    private static Mutex mutext { get; } = new (false, typeof(DisposeTests).FullName);
```
Keboo:
```suggestion
    private static Mutex Mutext { get; } = new (false, typeof(DisposeTests).FullName);
```

src/Chapter10.Tests/Listing10.17.RegisteringAFinalizerWithProcessExit.Tests.cs
```diff
@@ -19,25 +21,32 @@ public class DisposeTests
...
+        string psOutput;
         // Clean up at the start in case the class cleanup doesn't run (due to debug for example)
-        Assert.AreEqual<int>(0, RunPowerShellScript("cleanup", out string _));
-        string testStage = "create";
-        Assert.AreEqual<int>(0, RunPowerShellScript(testStage, out string psOutput),
+        string testStage = "cleanup";
+        Assert.AreEqual<int>(0, RunPowerShellScript(testStage, out psOutput),
+            $"Script failed with $testStage='{testStage}'. psOutput:\n{psOutput}");
+        testStage = "create";
+        int exitCode = RunPowerShellScript(testStage, out psOutput);
```
Keboo, line 34: Do these executions also need to acquire the mutex?

Keboo (approved):

src/Chapter10.Tests/Listing10.17.RegisteringAFinalizerWithProcessExit.Tests.cs
MarkMichaelis: They are in a class initializer so should only run once and before anything else in the test. In summary, I don't think synchronization is required.

Keboo: That is not quite accurate. We are probably fine for now because we are just taking the defaults for MSTest, however it supports method level parallelization. Each execution of a test method will create a new instance of the test class, and if two tests are executed in parallel it is possible to have two threads executing this method at the same time.

https://learn.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2022#mstest-element

PR #507, opened 2023-08-10
Switched to use nameof with parameter in method attribute.

MarkMichaelis:


BenjaminMichaelis (changes requested):
something seems weird here, Either the exludes can be moved up or something. Nothing looks "switched", just added

BenjaminMichaelis (approved):

PR #519, opened 2023-08-15
Added sample for Covariant Return Types (7.11)

MarkMichaelis:


src/Shared/Tests/CompilerAssert.cs
```diff
@@ -4,11 +4,36 @@
...
 
 namespace AddisonWesley.Michaelis.EssentialCSharp.Shared.Tests;
 
 public static class CompilerAssert
 {
+    /// <summary>
+    /// Compile the file associated with he caller's file name.
+    /// </summary>
+    /// <param name="expectedErrorIds"></param>
+    /// <returns></returns>
```
BenjaminMichaelis:
```suggestion
    /// <param name="expectedErrorIds">The id's of the expected errors and warnings. ex: CS0165</param>
    /// <returns>A task that is compiling the code.</returns>
```

Keboo (commented):
Did you mean to edit things across 4 chapters here? The title of this PR suggests only the chap 7 changes are expected.

MarkMichaelis:
@keboo: Sort of... I renumbered things in Chapter 7 and then did a multi-chapter change that modified Chapter 7. To avoid a nasty merge with multiple file changes and numerous file name changes, did one change on top of the other. There are separate branches if that is easier, but the one branch builds on the other.

src/Chapter07/Listing07.11.CovariantReturnTypes.cs
```diff
@@ -0,0 +1,14 @@
+#pragma warning disable IDE0059 // Unnecessary assignment of a value
```
Keboo, line 1: Is this pragma actually needed. I don't see any local value assignments occurring in this sample.

src/Shared/Tests/CompilerAssert.cs
```diff
@@ -4,11 +4,36 @@
...
+            GetTargetFileNameToCompileFromTestFileName(testFileName) }, 
+            expectedErrorIds);
+    }
+    
+    public static string GetTargetFileNameToCompileFromTestFileName(
+        [CallerFilePath] string testFileName = null!)
+    {
+#if NET7_0_OR_GREATER
+        ArgumentException.ThrowIfNullOrEmpty(testFileName);
+#endif
```
Keboo, line 31: Should there be an #else that also does the null check?

src/Shared/Tests/CompilerAssert.cs
```diff
@@ -35,14 +60,35 @@ public static async Task CompileAsync(string[] fileNames, string[] expectedError
...
+                    // Perhaps we have a wildcard in the file name.
+                    fileNamesToCompile.AddRange(Directory.EnumerateFiles(
+                        currentTargetDirectory,
+                        eachFileName,
+                        SearchOption.TopDirectoryOnly));
+                }
+                else
+                {
+                    throw new ArgumentException(
+                        $"Specified file, {eachFileName} does not exist.");
```
Keboo:
```suggestion
                    throw new ArgumentException(
                        $"Specified file, {eachFileName} does not exist.", nameof(fileNames));
```

Keboo (approved):

src/Shared/Tests/CompilerAssert.cs
MarkMichaelis: I decided not to do this because we would just have to remove when we drop .NET 6 , it and it wasn't germane anyway.

src/Chapter07/Listing07.12.OverrideVersusNewModifier.cs
```diff
@@ -1,5 +1,5 @@
 #pragma warning disable IDE0059 // Unnecessary assignment of a value
```
MarkMichaelis:
```suggestion
```

MarkMichaelis:
```suggestion
```

PR #532, opened 2023-08-29
Chapter 23 listing updates

Keboo:
## Summary

Updates to go with document changes.
Updated to show function pointers and inline arrays


src/Chapter23.Tests/Listing23.20.DesignatingABlockForUnsafeCode.Tests.cs
```diff
@@ -2,7 +2,7 @@
 using System.Runtime.InteropServices;
 
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter23.Listing23_20.Tests;
-
+#if NET8_0_OR_GREATER
```
Keboo, line 5: @MarkMichaelis how do you want to handle this example? I rev'd it up to latest stuff to show off the use of the new features. I could do something conditional so it works on the old TFMs, do we want to shows those too?

MarkMichaelis (approved):

PR #551, opened 2023-09-20
Re-enabled disabled web download tests from Chapter 20

MarkMichaelis:


MarkMichaelis:
The tests are already enabled in v12 via a different PR.

PR #552, opened 2023-09-20
Feat: Review Spelling of Listing Code

danOIntellitect:
## Summary

Added a spell check editor config along with a exlusion.dic file that contains the strings that we determine are spelled correctly. Then I went through the source code and corrected the misspelled words.

Fixes 1618 from the other repo : https://github.com/IntelliTect-dev/EssentialCSharp.Tooling/issues/1618


.editorconfig
```diff
@@ -0,0 +1,9 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
```
Keboo: nit, I would start with the default editorconfig. You can generate one by running `dotnet new editorconfig` from the root of the repo.

exclusion.dic
```diff
@@ -0,0 +1,68 @@
...
+csproj
+Usings
+warnaserror
+lpfl
+cpuid
+retq
+vmax
+decryptor
+Abidal
+Autoproperty
```
Keboo: Can we drop this and fix the usage to be "AutoProperty"?

exclusion.dic
```diff
@@ -0,0 +1,68 @@
...
+Dever
+Stokesbary
+Heavey
+enumtest
+Equatable
+Fireswamp
+spokane
+urls
+iassist
+worldvision
```
Keboo: Is it possible to drop this and fix usage to be `worldVision`?

exclusion.dic
```diff
@@ -0,0 +1,68 @@
...
+Deconstructors
+Hartfelt
+Lindherst
+Parksdale
+Monteray
+inigo
+Machava
+Uncompress
+firstname
+lastname
```
Keboo: Can we drop these in favor of change usage to have a capital N for name?

src/Chapter01/Listing01.22.SampleCILOutput.cs
```diff
@@ -58,7 +58,7 @@ .maxstack 8
     } // End of method System.Void HelloWorld.Program::Main(System.String[])
 
 
-  .method public hidebysig specialname rtspecialname instance void .ctor() cil managed
+  .method public hidebysig specialName rtspecialname instance void .ctor() cil managed
```
Keboo: Sorry I was wrong when I told you to make this change, I did not realize it was an CIL identifier. This should be all lowercase.

Keboo (approved):

exclusion.dic
danOIntellitect: image
Here is the usage, I would argue to keep it worldvision since it is a link. But capital V still works for the link

danOIntellitect: I made the change in the most recent commit, but did it separately if we want to roll it back

Keboo (approved):

PR #576, opened 2023-11-03
feat: Bump CompilerAssert to net8.0

BenjaminMichaelis:
Blocked until 8.0 officially releases I think

### DIFF
--- a/src/Shared/Tests/CompilerAssert.cs
+++ b/src/Shared/Tests/CompilerAssert.cs
@@ -41,11 +41,11 @@ public static class CompilerAssert
         {
             CompilerDiagnostics = CompilerDiagnostics.Warnings,
             ReferenceAssemblies = new ReferenceAssemblies(
-                        "net7.0",
+                        "net8.0",
                         new PackageIdentity(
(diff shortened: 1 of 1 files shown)


src/Shared/Tests/CompilerAssert.cs
```diff
@@ -91,7 +91,7 @@ public static async Task CompileAsync(string[] fileNames, string[] expectedError
 
         List<string> usingsToIgnore = new()
         {
-            "Microsoft.VisualStudio.TestTools.UnitTesting"
+            nameof(Microsoft.VisualStudio.TestTools.UnitTesting),
```
Keboo: This is not the same. This will only get you the string "UnitTesting"

Keboo (changes requested):

src/Shared/Tests/CompilerAssert.cs
BenjaminMichaelis: Not sure what I was thinking, thanks :)

Keboo (approved):

BenjaminMichaelis:
Closing in favor of #584

PR #580, opened 2023-11-04
feat: Add rule for modifier order and fix violations

### LINKED ISSUE #262: Refactor to access modifier, static, then return type
MicrosoftTeams-image (2)
In 6.36 we have static public string instead of public static string;

Which ordering do we want to make consistent? Right now typically I see is access specifier then static, then return value, then name.

BenjaminMichaelis:
Fixes #262

### DIFF
--- a/src/Chapter07/Listing07.31.SwitchWithPatternMatching.cs
+++ b/src/Chapter07/Listing07.31.SwitchWithPatternMatching.cs
@@ -27,7 +27,7 @@ public class Program
             compositeFormatString, date.Year, date.Month, date.Day) : null;
     #endregion INCLUDE
 }
-static file class DateDeconstructors
+file static class DateDeconstructors
 {
     public static void Deconstruct(this DateTime date, out int year, out int month, out int day) => (year, month, day) = (date.Year, date.Month, date.Day);
 }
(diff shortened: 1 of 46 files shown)


src/Chapter07/Listing07.31.SwitchWithPatternMatching.cs
```diff
@@ -27,7 +27,7 @@ input switch
             compositeFormatString, date.Year, date.Month, date.Day) : null;
     #endregion INCLUDE
 }
-file static class DateDeconstructors
+static file class DateDeconstructors
```
Keboo: This one surprises me since the docs say that `file` should come before `static`
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036#csharp_preferred_modifier_order

Keboo (approved):

src/Chapter07/Listing07.31.SwitchWithPatternMatching.cs
BenjaminMichaelis: I agree, I opened an issue and pr on Roslyn to see if this needs an update, going to wait to merge this until I hear back on that, as I might need to dial down the warning to a message if they have a mistake until it's patched. https://github.com/dotnet/roslyn/issues/70702

Keboo: Sounds good to me. I think for now I would suggest switching it back as I suspect that is the most likely default.

BenjaminMichaelis: Reverted!

BenjaminMichaelis: Gonna merge this for now. If something changes I'll revert

PR #584, opened 2023-11-05
feat: Get NetCore Preprocessor Directive, Reference Assembly, and Package Identity Dynamically

BenjaminMichaelis:


src/Shared/NetCore.cs
```diff
@@ -4,13 +4,20 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Shared;
 
 public static class NetCore
 {
+    private static readonly char[] separator = new[] { '/', '\\' };
```
Keboo:
```suggestion
    private static readonly char[] _Separators = new[] { '/', '\\' };
```

src/Shared/NetCore.cs
```diff
@@ -4,13 +4,20 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Shared;
...
         Assembly assembly = typeof(System.Runtime.GCSettings).GetTypeInfo().Assembly;
-        string[] assemblyPath = assembly.Location!.Split(new[] { '/', '\\' }, StringSplitOptions.RemoveEmptyEntries);
+        string[] assemblyPath = assembly.Location!.Split(separator, StringSplitOptions.RemoveEmptyEntries);
         int netCoreAppIndex = Array.IndexOf(assemblyPath, "Microsoft.NETCore.App");
-        if (netCoreAppIndex > 0 && netCoreAppIndex < assemblyPath.Length - 2)
-            return assemblyPath[netCoreAppIndex + 1];
-        throw new Exception("Unable to determine .NET Core version.");
+        return netCoreAppIndex > 0 && netCoreAppIndex < assemblyPath.Length - 2
+            ? assemblyPath[netCoreAppIndex + 1]
+            : throw new Exception("Unable to determine .NET Core version.");
```
Keboo:
```suggestion
            : throw new InvalidOperationException("Unable to determine .NET Core version.");
```

Keboo (approved):

src/Chapter06/Listing06.39.SpecifyingRequiredMembersWithinTheObjectInitializer.cs
```diff
@@ -1,27 +1,23 @@
...
+        #endif // COMPILEERROR // EXCLUDE
 
         // ...
         #endregion INCLUDE
-        #endif // COMPILEERROR // EXCLUDE
-      }
+    }
 }
-#endif // NET7_0_OR_GREATER
+#endif // NET7_0_OR_GREATER // EXCLUDE
```
Keboo, line 23: This looks a bit awkward to have dual `//` in the same line.

src/Shared/NetCore.cs
```diff
@@ -4,13 +4,48 @@ namespace AddisonWesley.Michaelis.EssentialCSharp.Shared;
...
+        }
+
+        // ex: NETCOREAPP3_1_OR_GREATER
+        for (int i = 1; i < 4; i++)
+        {
+            for (int j = 0; j <= 1; j++)
+            {
+                yield return $"NETCOREAPP{i}_{j}_OR_GREATER";
+            }
+        }
```
Keboo, line 49: This works, but I think it is cleaner, to just write out the individual `yield return` lines.

Keboo (approved):

src/Chapter06/Listing06.39.SpecifyingRequiredMembersWithinTheObjectInitializer.cs
BenjaminMichaelis: I think the parser right now looks for `// EXCLUDE` explicitly if I am reading the regex right https://github.com/IntelliTect-dev/EssentialCSharp.Tooling/blob/12dffde4ff4dfbc27f0c4cd6b49b8bad8bb15cbb/Parser/CSharp/CSharpParser.cs#L431-L432C14

src/Shared/NetCore.cs
BenjaminMichaelis: Are you suggesting something like:

```C#
yield return NETCOREAPP3_1_OR_GREATER;
yield return NETCOREAPP3_0_OR_GREATER;
etc...
```

Keboo: Yea exactly

src/Chapter06/Listing06.39.SpecifyingRequiredMembersWithinTheObjectInitializer.cs
Keboo: Yuck, yea you are correct, but this is now one more case where our "source" data is going to start to look awkward for the sake of the parser. I am fine leaving this as is, but still making me think about better ways to share the code.

src/Shared/NetCore.cs
BenjaminMichaelis: I'm going to push this to a separate issue #586 where I can focus on tackling this a bit better overall.

PR #590, opened 2023-11-08
feat: Switch == null and != null to is and is not null

### LINKED ISSUE #212: check to make sure == null is replaced with is null

BenjaminMichaelis:
Fixes #212


Keboo (approved):
Change looks fine. 
Two things though.
1. Did you check the content around these listing to see it needs to be updated as well?
2. Some current MS analyzers don't properly pick up `is null` as a null check. This may not be that big of a deal though.

BenjaminMichaelis:
> Change looks fine. Two things though.
> 
> 1. Did you check the content around these listing to see it needs to be updated as well?
> 2. Some current MS analyzers don't properly pick up `is null` as a null check. This may not be that big of a deal though.

1. Yes, content looks good, we usually say something generic like "making a null check", not "we use == null"
2. I assume this isn't an issue until we see it is? is that correct or should I worry about this before merging?

Keboo:
yep :shipit:

PR #596, opened 2023-11-13
feat: Write tests for compress and decompress methods

### LINKED ISSUE #20: Listing 18.15B (to become 18.16) - ValueTaskAsyncReturn: Improvements Needed
Create unit tests for Listing 18.15B (to become 18.16) and add a Main() method that calls the Compress/Uncompress methods (given a file name for args.).

BenjaminMichaelis:
Fixes #20

### DIFF
--- a/src/Chapter20/Listing20.04.ValueTaskAsyncReturn.cs
+++ b/src/Chapter20/Listing20.04.ValueTaskAsyncReturn.cs
@@ -1,8 +1,10 @@
 namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter20.Listing20_04;
 
+using System;
+
 #region INCLUDE
 using System.IO;
-using System.Text;
+using System.IO.Compression;
 using System.Threading.Tasks;
 
 public static class Program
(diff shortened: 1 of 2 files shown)


src/Chapter20/Listing20.04.ValueTaskAsyncReturn.cs
```diff
@@ -13,45 +15,30 @@ public static async ValueTask<byte[]> CompressAsync(byte[] buffer)
...
-        int bytesRead = gZipStream.Read(byteArray, 0, byteArray.Length);
-
-        //Transform byte[] unzip data to string
-        StringBuilder data = new(bytesRead);
-        //Read the number of bytes GZipStream red and do not a for each bytes in
-        //resultByteArray
-        for (int i = 0; i < bytesRead; i++)
+        using MemoryStream inputStream = new(buffer);
+        using MemoryStream uncompressedStream = new();
+        using (GZipStream gZipStream = new(inputStream, CompressionMode.Decompress))
```
Keboo, line 36:
```suggestion
        using GZipStream gZipStream = new(inputStream, CompressionMode.Decompress);
```

Keboo (approved):

src/Chapter20/Listing20.04.ValueTaskAsyncReturn.cs
BenjaminMichaelis: I can't do this because this will cause it to fail because the scoping will be different and dispose too late. It's a fun case that I might write a quick blog post about :)

Keboo: Interesting. I would be curious to understand that. Since it appears that the usage should be done.

BenjaminMichaelis: Happy to look at it without you if you want. Totally possible I'm missing something. Took me a long time to figure out that this caused an actual difference in the result though (the scoping of the using). I think because the pointer is at the end of the stream until it gets flushed? But I need to look closer

BenjaminMichaelis: Going to resolve this conversation and move this continued conversation off GitHub.

PR #790, opened 2024-08-28
feat: Update example of String Format instead of just round trip

### LINKED ISSUE #616: [Errata] Page 62
Based on C# 12.0 and .NET 8.0, in the "Advanced Topic: Round-Trip Formatting," both the example and the output are incorrect. The entire "Advanced Topic" needs to be rewritten.

BenjaminMichaelis:
Changes example to just show off String Format specifiers instead of just Round Trip format - this will support doc changes

Using CultureInfo as we are in an advanced block

References: 
- https://github.com/dotnet/docs/issues/41766
- https://github.com/IntelliTect/EssentialCSharp.Web/discussions/521


BenjaminMichaelis:
I was tossing back and forth the idea of adding another ToString with explicitly setting US format in case we have a reader from FR - but that seemed maybe overkill and adding clutter while not adding much value. Thoughts?

src/Chapter02/Listing02.09.UsingStringFormat.cs
```diff
@@ -0,0 +1,30 @@
...
+        Console.WriteLine(currencyText);
+
+        currencyText = $"{number:C}";
+        Console.WriteLine(currencyText);
+
+        // Prove that string interpolation and the ToString method produce equivalent results with format specifiers
+        string toStringCurrencyText = number.ToString("C");
+        Console.WriteLine($"{currencyText == toStringCurrencyText}: {currencyText} == {toStringCurrencyText}");
+
+        toStringCurrencyText = number.ToString("C", CultureInfo.GetCultureInfo("el-GR"));
```
Keboo, line 26: Can we add a comment indicating that this is greek?

Keboo (approved):

PR #794, opened 2024-11-06
Chapter13.15 bug fix

twoody0:
## Summary
This pull request addresses an issue in `Chapter13.15.ParameterlessStatementLambdas.cs` where `getUserInput` was only taking white space as input. The method has been updated to ensure that only valid, non-whitespace strings are accepted.

## Changes Made
- Removed the Not (`!`) on the while loop condition in `getUserInput` to properly handle whitespace inputs.
- Updated 'Main' to call `Console.WriteLine(input);` for testing.
- Added unit test to verify correct behavior for valid inputs.

### DIFF
--- a/src/Chapter13.Tests/Listing13.15.Tests.cs
+++ b/src/Chapter13.Tests/Listing13.15.Tests.cs
@@ -1,0 +1,31 @@
+﻿using System.IO;
+
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter13.Listing13_15.Tests;
+
+[TestClass]
+public class ProgramTests
+{
+    [TestMethod]
+    public void MainTest()
+    {
+        const string expectedInput = "Hello, World!";
+        const string expectedOutput = "Hello, World!";
+
+        StringReader reader = new (expectedInput);
+        StringWriter writer = new ();
+        try
+        {
+            Console.SetIn(reader);
+            Console.SetOut(writer);
+            Program.Main();
+
+            string actualOutput = writer.ToString().Trim();
+            Assert.AreEqual(expectedOutput, actualOutput);
+        }
+        finally
+        {
+            reader.Dispose();
+            writer.Dispose();
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Chapter13.Tests/Listing13.15.Tests.cs
```diff
@@ -0,0 +1,30 @@
+﻿
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter13.Listing13_15.Tests;
+
+[TestClass]
+public class ProgramTests
+{
+    [TestMethod]
+    public void MainTest()
```
Keboo, line 7: This is great, but we have a built in tool for doing console assertions that can make this a bit easier. If you look [here](https://github.com/IntelliTect/EssentialCSharp/blob/v12.0/src/Chapter13.Tests/Listing13.12.Tests.cs) (other tests as well), you can see the syntax for leveraging the ConsoleAssert class to be able to test the input/output from the console.

twoody0: I edited the test to use the ConsoleAssert class

src/Chapter13.Tests/Listing13.15.Tests.cs
```diff
@@ -0,0 +1,24 @@
+﻿
```
BenjaminMichaelis:
```suggestion
```
nit: whitespace

src/Chapter13.Tests/Listing13.15.Tests.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace AddisonWesley.Michaelis.EssentialCSharp.Chapter13.Listing13_15.Tests;
+
+[TestClass]
+public class ProgramTests
+{
+    [TestMethod]
+    public void MainTest()
+    {
+        const string expectedOutput = "ValidInput";
```
BenjaminMichaelis:
```suggestion
        const string expectedOutput = ""<<ValidInput>>ValidInput"";
```
TestTools.Console should be able to handle both input and output: https://github.com/IntelliTect/TestTools.Console?tab=readme-ov-file#basic-usage

This would also all you to simplify the Expect call to just need the expected string and the `Action`

twoody0: I believe I implemented it how you were expecting, that's much simpler than what I was trying to do. Thank you!

BenjaminMichaelis: I updated the errata to include your contribution and give you credit and pushed the change to your branch if you would like to make sure it looks ok

twoody0: Perfect!

BenjaminMichaelis (approved):

Keboo (approved):

